PR #147, opened 2023-03-01
Synthetic noise psd dataset pr

jonaswildberger:
Re-factored version of the addition of the synthetic PSD dataset generation in https://github.com/dingo-gw/dingo-devel/pull/128. Now, the main routine accepts an existing `asd_dataset.hdf5` file, parameterizes all ASDs in there, and re-samples a new dataset. Also added a short section in the documentation.


docs/source/noise_dataset.ipynb
```diff
@@ -42,47 +42,73 @@
...
     "  detectors:\n",
     "    - H1\n",
     "    - L1\n",
     "  observing_run: O2\n",
+    "condor:\n",
+    "  env_path: path/to/environment\n",
+    "  num_jobs: 2    # per detector\n",
+    "  num_cpus: 16\n",
+    "  memory_cpus: 16000\n",
+    "  bid: 200\n",
```
stephengreen, line 97: Does the `bid` ever get used? I couldn't find it in the code. Also this is specific to the MPI-IS cluster, so it should be optional.

dingo/gw/noise/utils.py
```diff
@@ -0,0 +1,245 @@
...
+import requests
+import yaml
+
+from gwpy.table import EventTable
+from dingo.gw.noise.asd_dataset import ASDDataset
+
+"""
+Catalogue against which to check that no event is present in the estimated PSDs
+"""
+CATALOGUEs = ["GWTC-1-confident", "GWTC-2.1-confident", "GWTC-3-confident"]
```
stephengreen, line 20: It's a bit awkward with the small "s", and probably better to use the American spelling, CATALOGS.

dingo/gw/noise/utils.py
```diff
@@ -0,0 +1,245 @@
...
+        type=str,
+        required=True,
+        help="Path where the PSD data is to be stored. Must contain a 'settings.yaml' file.",
+    )
+    parser.add_argument(
+        "--settings_file",
+        type=str,
+        required=True,
+        help="Path to a settings file in case two different datasets are generated in the same directory",
+    )
```
stephengreen, line 192: Same issue as above: the settings file is not optional, so why should data-dir contain a settings file?

dingo/gw/noise/utils.py
stephengreen, line 1: Supposing we are at the end of O4, is there a way to download O4 data? This seems to only work using the pre-specified set of paths. This doesn't have to be fixed now, but later it should be changed.

dingo/gw/noise/synthetic_noise/generate_dataset.py
```diff
@@ -0,0 +1,117 @@
...
+from dingo.gw.noise.synthetic_noise.asd_sampling import KDE, get_rescaling_params
+from dingo.gw.noise.synthetic_noise.utils import reconstruct_psds_from_parameters
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=textwrap.dedent(
+            """\
+        Generate a synthetic noise ASD dataset from an existing dataset of real ASDs.
```
stephengreen, line 21: This help should be a bit more complete: The synthetic ASD dataset also depends on an additional ASD for scaling, and it's not clear that this is the case, or how this should be specified.

dingo/gw/noise/synthetic_noise/generate_dataset.py
```diff
@@ -0,0 +1,117 @@
...
+        num_processes,
+        verbose=verbose,
+    )
+
+    synthetic_dataset_dict = copy.deepcopy(real_dataset.to_dictionary())
+    synthetic_dataset_dict["settings"]["parameterization_settings"] = settings[
+        "parameterization_settings"
+    ]
+
+    sampling_settings = settings.get("sampling_settings", None)
```
stephengreen, line 67: The default second argument to `dict.get` is `None`.

```suggestion
    sampling_settings = settings.get("sampling_settings")
```

dingo/gw/noise/synthetic_noise/generate_dataset.py
```diff
@@ -0,0 +1,117 @@
...
+            psds[:, real_dataset.domain.min_idx : real_dataset.domain.max_idx + 1]
+        )
+
+    synthetic_dataset_dict["parameters"] = parameters_dict
+    synthetic_dataset_dict["asds"] = asds_dict
+    # TODO: should I make the "data_keys" an optional argument to the ASDDataset class? Such that the parameters
+    # can also be stored in this way
+    return DingoDataset(
+        dictionary=synthetic_dataset_dict, data_keys=["asds", "gps_times", "parameters"]
+    )
```
stephengreen, line 98: Is there a reason you don't use an `ASDDataset` here?

dingo/gw/noise/synthetic_noise/asd_sampling.py
```diff
@@ -0,0 +1,112 @@
...
+            split_indices = sorted(split_indices)
+
+            for i in range(len(split_indices) - 1):
+                vals = y_values[:, split_indices[i]: split_indices[i + 1]].T
+                kde_vals = stats.gaussian_kde(
+                    vals, bw_method=float(self.settings["bandwidth_spline"])
+                )
+                self.broadband_kde[det].append(kde_vals)
+
+    def sample(self, rescaling_ys):
```
stephengreen, line 64: Should `rescaling_ys` have a default argument of `None`, for the case where we do not want to rescale the sampled PSDs? The code seems to accept `None`, but it should also probably be the default.

I think it would be useful for this method to also have a docstring.

dingo/gw/noise/synthetic_noise/asd_sampling.py
```diff
@@ -0,0 +1,112 @@
...
+                y_values_mean = np.mean(y_values, axis=0)
+                y_values = (
+                    y_values - y_values_mean[None, :] + rescaling_ys[det]
+                )
+            parameters_dicts[det]["y_values"] = y_values
+
+        return parameters_dicts
+
+
+def get_rescaling_params(filenames, parameterization_settings):
```
stephengreen, line 97: A docstring would be good here. It's not clear, e.g., what the `filenames` should be (i.e., ASDDataset files).

dingo/gw/noise/synthetic_noise/generate_dataset.py
```diff
@@ -0,0 +1,117 @@
...
+    sampling_settings = settings.get("sampling_settings", None)
+    if sampling_settings:
+        kde = KDE(parameters_dict, sampling_settings)
+        kde.fit()
+        rescaling_params = None
+        if "rescaling_psd_paths" in sampling_settings:
+            rescaling_params = get_rescaling_params(
+                sampling_settings["rescaling_psd_paths"],
+                settings["parameterization_settings"],
+            )
```
stephengreen, line 76: The variable names are confusing here. What you are calling `rescaling_psd_paths` should actually be paths to `ASDDataset` files, according to the code of `get_rescaling_params`. It would be best to rename this to `rescaling_asd_paths`.

Also, it should be made clear somewhere that the rescaling ASD is the *first* ASD in each dataset.

dingo/gw/noise/synthetic_noise/generate_dataset.py
stephengreen: Also this preserves the GPS times from the ASDDataset that you started from, even though the individual synthetic ASDs have nothing to do with the original ones. It would be better to drop the "gps_times" key from the dict, and make sure the `ASDDataset` is robust to this key missing.

dingo/gw/noise/synthetic_noise/asd_sampling.py
```diff
@@ -0,0 +1,112 @@
...
+            for i in range(len(split_indices) - 1):
+                vals = y_values[:, split_indices[i]: split_indices[i + 1]].T
+                kde_vals = stats.gaussian_kde(
+                    vals, bw_method=float(self.settings["bandwidth_spline"])
+                )
+                self.broadband_kde[det].append(kde_vals)
+
+    def sample(self, rescaling_ys):
+
+        num_samples = self.settings["num_samples"]
```
stephengreen, line 66: Should this maybe be an argument to `sample`?

dingo/gw/noise/synthetic_noise/asd_parameterization.py
```diff
@@ -0,0 +1,219 @@
...
+def fit_broadband_noise(domain, psd, num_spline_positions, sigma):
+    frequencies = domain.sample_frequencies
+
+    # standardize frequencies to the interval [0,1]
+    standardized_frequencies = (frequencies - frequencies[0]) / (
+        frequencies[-1] - frequencies[0]
+    )
+
+    # log-distributed x-positions in [20/f_max, 1] for the interpolating base noise spline
+    log_xs = np.logspace(np.log10(F_MIN / domain.f_max), 0, num_spline_positions)
```
stephengreen, line 103: What is the reason to hard-code `F_MIN=20`? What if we want to analyze data going below 20 Hz?

dingo/gw/noise/synthetic_noise/asd_parameterization.py
```diff
@@ -0,0 +1,219 @@
...
+from threadpoolctl import threadpool_limits
+
+from dingo.gw.noise.synthetic_noise.utils import get_index_for_elem, lorentzian_eval
+
+F_MIN = 20
+
+P0_A, P0_Q = 5, 100
+MIN_A, MIN_Q = 0, 10
+MAX_A, MAX_Q = 12, 1000
+MAXFEV = 5000000
```
stephengreen, line 17: How were these values chosen? Experimentation? Or a paper?

stephengreen (changes requested):
Thank you Jonas! It's a big PR, so it took a while to review. There are a number of minor comments sprinkled throughout, but it looks great overall.

A few other suggestions:
* Add a configuration file to build a synthetic ASD dataset. Otherwise it's not clear how to do this.
* Rename the package `synthetic_noise` -> `synthetic`. It's already a sub-package of `noise`, so it does not need to repeat that.

dingo/gw/noise/synthetic_noise/asd_parameterization.py
jonaswildberger: experimentation

dingo/gw/noise/synthetic_noise/asd_parameterization.py
jonaswildberger: This needs to be >0 due to the log distribution, and usually the ASD is very unstable in the low frequency regions. For convenience, we just picked 20 at some point for our applications. The ASDs will still include values below 20, but these won't be fitted very accurately. I agree that this could be generalised at some point

dingo/gw/noise/synthetic_noise/asd_sampling.py
jonaswildberger: Yeah, I totally agree

dingo/gw/noise/synthetic_noise/generate_dataset.py
jonaswildberger: I cannot overwrite the `data_keys` of ASDDataset. Using ASDDataset wouldn't allow me to save `parameters`, see the comment in the code. Regarding the gps_times, this is intentional, since I thought it would be good to know data from which time window I've accessed to created the synthetic dataset, in particular if the PSDs were only parameterised and not re-sampled

stephengreen: Ah I see, you are saving the parameters as well. I guess when this gets loaded later by an ASDDataset, it just ignores the parameters. The only issue I foresee is that the `dataset_type` would not be set to `"asd_dataset"`, which would affect the output of `dingo_ls`. You could alternatively add `"parameters"` to the ASDDatset `data_keys`. (To avoid confusion, it might be better to call it `"asd_parametrizations"` or something like that.

examples/dataset_generation/asd_dataset_settings_synthetic.yaml
```diff
@@ -0,0 +1,15 @@
...
+  smoothen: True
+sampling_settings:
+   bandwidth_spectral: 0.5
+   bandwidth_spline: 0.25
+   split_frequencies:
+     - 30
+     - 100
+   rescaling_asd_paths:
+     H1: /path/to/rescaling_asd_H1.hdf5  # we use the first element here, in case the datasets contains multiple
+     L1: /path/to/rescaling_asd_L1.hdf5 # we use the first element here, in case the datasets contains multiple
```
stephengreen, line 15: Just fixing indentation.

```suggestion
  bandwidth_spectral: 0.5
  bandwidth_spline: 0.25
  split_frequencies:
    - 30
    - 100
  rescaling_asd_paths:
    H1: /path/to/rescaling_asd_H1.hdf5  # we use the first element here, in case the datasets contains multiple
    L1: /path/to/rescaling_asd_L1.hdf5 # we use the first element here, in case the datasets contains multiple
    
```

docs/source/noise_dataset.ipynb
```diff
@@ -129,6 +129,72 @@
...
+    ": Maximum number of spectral lines to model.\n",
+    "\n",
+    "`sigma`\n",
+    ": Standard deviation of the Normal distribution parameterizing $p(\\log S_n|z)$.\n",
+    "\n",
+    "`delta_f`\n",
+    ": If > 0, truncates each spectral line.\n",
+    "\n",
+    "`smoothen`\n",
+    ": Whether to save the smooth ASDs (True) or the noisy ASDs (False). The noisy synthetic ASDs resemble real ASDs estimated with Welch's method more closely, while the smooth ASDs are more similar to ASDs generated with BayesWave.\n",
```
stephengreen, line 184:
```suggestion
    ": Whether to save the smooth ASDs (True) or the noisy ASDs (False). The noisy synthetic ASDs resemble real ASDs estimated with Welch's method more closely, while the smooth ASDs are more similar to ASDs generated with BayesWave. (Default: False) \n",
```

docs/source/noise_dataset.ipynb
```diff
@@ -91,12 +117,36 @@
...
     ": Which detectors (H1, L1, V1, ...) to include in the dataset.\n",
     "\n",
     "Observing run\n",
     ": Which observing run to use when estimating PSDs.\n",
     "\n",
+    "Condor (optional)\n",
+    ": Settings for [HTCondor](https://htcondor.readthedocs.io/en/latest/index.html) useful for parallelizing the ASD estimation across condor jobs.\n",
+    "\n",
+    "### `dingo_generate_synthetic_asd_dataset`\n",
+    "This method generates a dataset of synthetic ASDs from a dataset of existing ASDs to enhance robustness against ASD distribution shifts. The implementation follows the steps explained in this [paper](https://inspirehep.net/literature/2182788).\n",
```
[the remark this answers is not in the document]
stephengreen:
```suggestion
    "This method generates a dataset of synthetic ASDs from a dataset of existing ASDs to enhance robustness against ASD distribution shifts. In particular, this allows to generate a dataset of synthetic ASDs that are *scaled* by a fiducial ASD in order to adapt to a new observing run. This is particularly useful for training Dingo networks at the beginning of an observing run, when the number of training ASDs is limited. The implementation follows the steps explained in this [paper](https://inspirehep.net/literature/2182788).\n",
```

docs/source/noise_dataset.ipynb
```diff
@@ -129,6 +129,72 @@
...
+    ": Bandwidths for the KDEs modeling the distribution over spectral lines and broad-band noise, respectively. These determine the width of the resulting distribution.\n",
+    "\n",
+    "`num_samples`\n",
+    ": Number of synthetic ASDs to generate.\n",
+    "\n",
+    "`split_frequencies`\n",
+    ": (Set of) frequencies at dividing the broad-band noise into independent segments, e.g. due to different dominant noise sources (shot noise, seismic noise, etc.).\n",
+    "\n",
+    "`rescaling_psd_paths`\n",
+    ": Paths to ASD datasets for each detector to which the synthetic ASDs should be rescaled, e.g. the PSDs from the target observing run. If the dataset contains multiple ASDs, we use the first one.\n",
```
stephengreen, line 196:
```suggestion
    ": Paths to ASD datasets for each detector to which the synthetic ASDs should be rescaled, e.g. the PSDs from the target observing run. If the dataset contains multiple ASDs, we use the first one. (Optional; if not provided, no rescaling will be done.)\n",
```

docs/source/noise_dataset.ipynb
```diff
@@ -91,12 +117,36 @@
     "`num_psds_max` (optional)\n",
     ": If set, stop building the dataset after this number of PSDs have been estimated. This setting is useful for building a single-PSD dataset for pretraining a network.\n",
     "\n",
+    "Channels 'channels (optional)\n",
```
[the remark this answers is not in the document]
stephengreen: I'm not sure why, but it looks like this got lost somehow.

docs/source/noise_dataset.ipynb
```diff
@@ -42,47 +42,73 @@
...
-    "Data length `T_PSD` (s)\n",
+    "Data length `time_psd` (s)\n",
     ": The entire length of data from which to estimate a PSD using Welch's method. Periodigrams are calculated on segments of this, and then averaged using the `median` method.\n",
     "\n",
     "Segment length `T` (s)\n",
     ": The length of each segment on which to take the DFT and calculate a periodigram.\n",
     "\n",
-    "Gap `T_gap` (s)\n",
+    "Gap `time_gap` (s)\n",
     ": Gap between duration-`T` segments. E.g., if `T_PSD=1024`, `T=8`, `T_gap=8`, then for each PSD, 64 periodigrams are computed, each using data stretches 8 s long, with gaps of 8 s between segments. Segments would then be $[0~\\text{s}, 8~\\text{s}], [16~\\text{s}, 24~\\text{s}], \\ldots$.\n",
```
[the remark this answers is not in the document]
stephengreen: Not sure why this shows as outdated.

docs/source/noise_dataset.ipynb
```diff
@@ -42,47 +42,73 @@
     "\n",
     "As with the noise realizations, a random ASD is chosen from the dataset when preparing each sample during training. This augments the training set compared to fixing the noise ASD for each sample prior to training.\n",
     "\n",
-    "Similarly to the `WaveformDataset`, the `ASDDataset` is just a container. Dingo includes routines for building such a dataset from observational data. The basic approach is as follows:\n",
+    "Similarly to the `WaveformDataset`, the `ASDDataset` is just a container. Dingo includes routines for building such a dataset from observational data.\n",
     "\n",
-    "1. Identify stretches of data within an observing run meeting certain criteria (sufficiently long, without events, and sufficiently high quality, ...).\n",
-    "2. Download data from [GWOSC](https://www.gw-openscience.org). \n",
+    "## Command-line scripts\n",
```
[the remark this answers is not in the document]
stephengreen: Same, still a suggestion.

docs/source/noise_dataset.ipynb
jonaswildberger: True, I don't think it does anymore

dingo/gw/noise/utils.py
jonaswildberger: The links are only needed for downloading from GWOSC. You can still download more up-to-date data by specifying the channels in the settings file

dingo/gw/noise/synthetic_noise/asd_parameterization.py
stephengreen: Okay, we should note to extend this, since at some point we may want to go below 20 Hz. Could you create an Issue so we can remember?

examples/dataset_generation/asd_dataset_settings.yaml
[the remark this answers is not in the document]
stephengreen: Your examples are now in a sufficiently separate directory, so there's no immediate conflict. But some settings names should be updated in #173.

docs/source/noise_dataset.ipynb
```diff
@@ -129,6 +128,72 @@
     "Condor (optional)\n",
     ": Settings for [HTCondor](https://htcondor.readthedocs.io/en/latest/index.html) useful for parallelizing the ASD estimation across condor jobs.\n",
     "\n",
+    "### `dingo_generate_synthetic_asd_dataset`\n",
+    "This method generates a dataset of synthetic ASDs from a dataset of existing ASDs to enhance robustness against ASD distribution shifts. In particular, it can be used to augment the training distribution when insufficient data is available, e.g. at the beginning of a new observing run or for BayesWave-like ASDs. The implementation follows the steps explained in this [paper](https://inspirehep.net/literature/2182788).\n",
```
stephengreen, line 132:
```suggestion
    "This method generates a dataset of synthetic ASDs from a dataset of existing ASDs to enhance robustness against ASD distribution shifts. In particular, this allows to generate a dataset of synthetic ASDs that are *scaled* by a fiducial ASD in order to adapt to a new observing run. This is particularly useful for training Dingo networks at the beginning of an observing run, when the number of training ASDs is limited. It also allows to generate smoother synthetic ASDs that more closely resemble those from BayesWave. The implementation follows the steps explained in this [paper](https://inspirehep.net/literature/2182788).\n",
```

dingo/gw/noise/synthetic_noise/generate_dataset.py
stephengreen: If it's not too much work, I would suggest it.

dingo/gw/noise/synthetic_noise/asd_parameterization.py
jonaswildberger: https://github.com/dingo-gw/dingo/issues/182

dingo/gw/noise/synthetic/generate_dataset.py
```diff
@@ -113,11 +113,11 @@ def generate_dataset(real_dataset, settings: Dict, num_samples, num_processes: i
             psds[:, real_dataset.domain.min_idx : real_dataset.domain.max_idx + 1]
         )
 
-    synthetic_dataset_dict["parameters"] = parameters_dict
+    synthetic_dataset_dict["asd_parameterizations"] = parameters_dict
     synthetic_dataset_dict["asds"] = asds_dict
```
stephengreen, line 118: Does this dict contain the GPS times here? If so, are they not meaningless at this point? Should they be del'd? (I don't know if they would cause problems with the ASDDataset class, so it's up to you.

```suggestion
        del synthetic_dataset_dict["gps_times"]

```

stephengreen (approved):
Looks great, thank you! I just put one more code suggestion, which you can take if you agree.

dingo/gw/noise/synthetic/generate_dataset.py
jonaswildberger: Yeah, it's copied from the initial dataset. I don't think they are generally meaningless, since 
a) we might only parameterise/smoothen the PSDs, in which case the gps_times still correspond to the psds and b) it may still be good to know what data was used for generating a synthetic dataset. But if you insist, I can delete it

PR #167, opened 2023-03-24
Add calibration uncertainty marginalization to dingo-pipe

stephengreen:
This pull request adds the ability to marginalize over detector calibration uncertainty to dingo-pipe. This takes place during the importance sampling stage.

It adds the following keys to the INI files:

* calibration-model
* spline-calibration-envelope-dict
* spline-calibration-nodes
* spline-calibration-curves (not present in bilby-pipe)

Usage is explained in the documentation. It should be noted that these have slightly different interpretation than in bilby-pipe, where one typically performs inference over the calibration parameters rather than marginalizing over them.

In addition, there was a small refactoring of the calibration marginalization code to simplify it.

@jonaswildberger You probably need to update #131 to accommodate these additional INI keys.

### DIFF
--- a/dingo/gw/pipe/importance_sampling.py
+++ b/dingo/gw/pipe/importance_sampling.py
@@ -117,6 +117,12 @@ class ImportanceSamplingInput(Input):
                 "num_calibration_nodes": self.spline_calibration_nodes,
                 "num_calibration_curves": self.spline_calibration_curves,
             }
+        elif self.calibration_model == None:
+            return None
+        else:
+            raise ValueError(
+                "The only calibration model which is supported is 'CubicSpline'"
+            )
 
     @property
     def importance_sampling_settings(self):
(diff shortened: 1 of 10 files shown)


dingo/gw/pipe/importance_sampling.py
```diff
@@ -108,6 +109,15 @@ def _load_event(self):
         event_dataset = EventDataset(file_name=self.event_data_file)
         self.result.reset_event(event_dataset)
 
+    @property
+    def calibration_marginalization_kwargs(self):
+        if self.calibration_model == "CubicSpline":
```
nihargupte-ph, line 114: I guess we should add an else here that just raises an exception.

nihargupte-ph:
Looks good overall! I will test things out a bit more tomorrow. Saraswati seems to not be able to ping github.com at the moment...

dingo/gw/pipe/importance_sampling.py
stephengreen: If it's `None` we just want it to do nothing. For any other string I suppose an exception makes sense.

stephengreen: Actually there'd be an error via the argument parser, which only accept `None` or "CubicSpline" at this point.

nihargupte-ph (approved):
Considering bilby has the same difference in evidence between using calibration marginalization and not using calibration marginalization. And considering the marginals are mostly unchanged when applying calibration marginalization. The current changes also match with my previous cal marg runs. I think things are working, thanks for the updates!

Issue #166, opened 2023-03-24
Replace chainconsumer with corner


stephengreen:
We use [chainconsumer](https://samreay.github.io/ChainConsumer/) within dingo for diagnostic corner plots, however this package is not present on conda-forge (and hence IGWN). I therefore suggest that we replace chainconsumer by [corner](https://corner.readthedocs.io/en/latest/) for this plotting.

PR #169, opened 2023-03-27
Replace chainconsumer with corner

stephengreen:
This replaces chainconsumer with corner for making corner plot, as described in #166. It places a multiple-posterior corner plotting function in `dingo.core.utils.plotting`.

I have tested it in `dingo_pipe`, but not in `inference_pipeline.py` and `diagnostics.py`. Could someone who uses these please test it?

### DIFF
--- a/dingo/core/utils/plotting.py
+++ b/dingo/core/utils/plotting.py
@@ -53,13 +53,21 @@ def plot_corner_multi(
     if not isinstance(labels, list):
         labels = [labels]
 
+    # Only plot common parameters for all sample sets, keeping the same order as the
+    # first sample set.
+    common_parameters = [
+        p
+        for p in samples[0].columns
+        if p in set.intersection(*(set(s.columns) for s in samples))
+    ]
+
     fig = None
     handles = []
     for i, (s, w, l) in enumerate(zip_longest(samples, weights, labels)):
         color = mpl.colors.rgb2hex(plt.get_cmap(cmap)(i))
         fig = corner.corner(
-            s.to_numpy(),
-            labels=s.columns,
+            s[common_parameters].to_numpy(),
+            labels=common_parameters,
             weights=w,
             color=color,
             no_fill_contours=True,
--- a/dingo/core/utils/plotting.py
+++ b/dingo/core/utils/plotting.py
@@ -70,16 +70,20 @@ def plot_corner_multi(
             labels=common_parameters,
             weights=w,
             color=color,
-            no_fill_contours=True,
+            no_fill_contours=False,
             fig=fig,
-            **corner_params
+            **corner_params,
         )
         handles.append(
             plt.Line2D([], [], color=color, label=l, linewidth=5, markersize=20)
         )
 
     # Eliminate spacing between the 2D plots
-    fig.subplots_adjust(wspace=0, hspace=0)
+    if len(common_parameters) > 8:
+        fig.subplots_adjust(wspace=0, hspace=0)
+    else:
+        space = 1 / (4 * len(common_parameters))
+        fig.subplots_adjust(wspace=space, hspace=space)
 
     fig.legend(
         handles=handles,
@@ -114,3 +118,5 @@ def plot_corner_multi(
     # Reset rcParams to original values
     mpl.rcParams["font.family"] = serif_old
     mpl.rcParams["lines.linewidth"] = linewidth_old
+
+    return fig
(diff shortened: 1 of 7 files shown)


dingo/core/utils/plotting.py
```diff
@@ -0,0 +1,108 @@
...
+    fig = None
+    handles = []
+    for i, (s, w, l) in enumerate(zip_longest(samples, weights, labels)):
+        color = mpl.colors.rgb2hex(plt.get_cmap(cmap)(i))
+        fig = corner.corner(
+            s.to_numpy(),
+            labels=s.columns,
+            weights=w,
+            color=color,
+            no_fill_contours=True,
```
nihargupte-ph, line 73: Maybe we want to allow for filled contours here, should this go in corner_params? Or maybe this only goes to hist_2d? When I set it to False nothing happens anyway.

dingo/core/utils/plotting.py
```diff
@@ -0,0 +1,108 @@
...
+            fig=fig,
+            **corner_params
+        )
+        handles.append(
+            plt.Line2D([], [], color=color, label=l, linewidth=5, markersize=20)
+        )
+
+    # Eliminate spacing between the 2D plots
+    fig.subplots_adjust(wspace=0, hspace=0)
```
nihargupte-ph, line 82: I guess this is up to the user, should it be 0? Sometimes this makes the posteriors a bit hard to interpret I attach a photo here; 
image

Vs when we were using chainconumser: 
image

dingo/core/utils/plotting.py
stephengreen: I experimented with filled contours, and I did not like how they looked unfortunately. I tried to make them semi-transparent as we currently have them, using `contourf_kwargs = {"alpha": 0.3}`, but this did not work. (There seems to be a bug, see https://github.com/dfm/corner.py/issues/112).

If you can get it working, please feel free to do so.

dingo/core/utils/plotting.py
stephengreen: I think ChainConsumer reduces the separation to 0 when we have more than a certain number of parameters. I just set this for all cases, but if you want to base it on `len(common_parameters)` feel free to do so.

dingo/core/utils/plotting.py
stephengreen: Also, to turn on the filling (aside from setting alpha) you need to remove the `no_fill_contours` argument, and set `fill_contours=True`.

nihargupte-ph:
Nice work, I think it looks reallly good overall! Getting all those corner plots to work must have been hard! 

I don't think there is a KDE option for the marginals in corner, but at some point when I used corner I wrote a script to automatically do this. But chainconsumer has this by default. That being said I don't know if it's worth it to do at the moment since we want to start testing on conda-forge ASAP.

stephengreen:
I actually asked GPT-4 how to fully smooth the 1D marginals, and it wrote a code to plot those axes with seaborn. This seemed excessive; I think we can live with the less smooth ones.

dingo/core/utils/plotting.py
nihargupte-ph: Ah, I see there is that bug.

For now maybe I can just set the default to false in ```corner_params``` and then the user can update it w/ ```corner_params.update(kwargs)```. How does it sound?

stephengreen: Yeah, the user can update the kwargs as it stands now. But for filling the contours we may want to do this selectively for each sample set (especially with alpha = 1).

I also left one of the old ChainConsumer plotting functions, which now imports the package in a try/except block. So it's still there. In fact, one could set Dingo to use ChainConsumer in all cases when it is available. But that is a bit more work.

nihargupte-ph:
If GPT-4 can't do it there's no chance for us :)

I think what I was doing was individually taking the 1d and 2d marginals and training KDEs on them, and then resampling the KDEs. This in general does take a long time to create the plot. But to be honest, I agree the less smooth ones are fine. Maybe something to implement down the line if there is time.

dingo/core/utils/plotting.py
nihargupte-ph: Sounds good!

stephengreen:
> Ah also, should we be returnning the figure object? I think it could be quite useful if the user wants to edit the figure or add things on later?

Fine with me. In general I wasn't trying to leave a huge amount of flexibility to the user, but if you think this is useful please go ahead.

nihargupte-ph:
Ah right, I mean to change that back to how it was before. I can fix it when I'm back to my laptop.

PR #170, opened 2023-03-28
Model weights as hdf5

mpuerrer:
Implement conversion and loading of trained models to HDF5 format for inference

- Use the script `pt_to_hdf5.py` to convert trained models to HDF5 format. Entry point: `dingo_pt_to_hdf5`
- For inference simply specify the `.hdf5` model files.
- Training is not supported because optimizer_state_dict is not saved in the HDF5 file which saves a significant amount of disk space.
- The specified version number is added to the output filename (HDF5), i.e. `model_v1.hdf5`.
- Added required HDF5 attribute for CVMFS: `CANONICAL_FILE_BASENAME` is set to the basename of the hdf5 file. These need to be in sync which is checked when reading an HDF5 network.
- Add `fletcher32` filter when creating a HDF5 dataset where possible (except for scalar datasets): see https://docs.h5py.org/en/stable/high/dataset.html#fletcher32-filter


mpuerrer:
See the PR at https://github.com/dingo-gw/dingo/pull/170 instead.

dingo/core/models/posterior_model.py
```diff
@@ -185,6 +190,47 @@ def save_model(
...
         torch.save(model_dict, model_filename)
 
+
+    def _load_model_from_hdf5(
+        self,
+        model_filename: str
+    ):
+        """
+        Helper function to load a trained model that has been
+        saved in HDF5 format using `pt_to_hdf5.py`.
```
stephengreen, line 200:
```suggestion
        saved in HDF5 format using `dingo_pt_to_hdf5`.
```

dingo/core/models/posterior_model.py
```diff
@@ -185,6 +190,47 @@ def save_model(
...
+            if fp.attrs['CANONICAL_FILE_BASENAME'] != model_basename:
+                raise ValueError('HDF5 attribute CANONICAL_FILE_BASENAME differs from model name',
+                        model_basename)
+
+            # Load small nested dicts from json
+            for k, v in fp['serialized_dicts'].items():
+                d[k] = json.loads(v[()])
+
+            # Load model weights
+            model_state_dict = OrderedDict()
```
stephengreen, line 226: Why use an `OrderedDict` here? `dict` objects are already ordered now in Python.

dingo/core/utils/pt_to_hdf5.py
```diff
@@ -0,0 +1,77 @@
...
+import torch
+import numpy as np
+import h5py
+import json
+import argparse
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Convert the weights of a trained PyTorch model to HDF5 format"
```
stephengreen, line 11:
```suggestion
        description="Convert the weights of a trained Dingo model from a PyTorch pickle .pt file to HDF5,"
```

dingo/core/utils/pt_to_hdf5.py
```diff
@@ -0,0 +1,77 @@
...
+        " for distribution in the LVK's CVMFS.",
+        epilog="Training history (optimizer_state_dict) is discarded.")
+    parser.add_argument("-i", "--in_file", type=str, required=True,
+            help='Input model ".pt" weights file')
+    parser.add_argument("-o", "--out_file", type=str, required=True,
+            help='Output model ".hdf5" weights file')
+    parser.add_argument("-n", "--version_number", type=int, required=True,
```
stephengreen, line 18: I suggest to specify "model" version to avoid confusion with Dingo version.
```suggestion
    parser.add_argument("-n", "--model_version_number", type=int, required=True,
```

dingo/core/utils/pt_to_hdf5.py
```diff
@@ -0,0 +1,77 @@
...
+    out_file_name = f'{root}_v{version_number}{ext}'
+    print('Output will be written to', out_file_name)
+
+    # Load data into CPU memory since we'll be saving it using CPU libraries
+    d = torch.load(args.in_file, map_location=torch.device("cpu"))
+
+    # Collect the names of the dicts that can be serialized to JSON; model_state_dict and
+    # optimizer_state_dict contain torch.tensors and cannot be JSON serialized
+    dicts_to_serialize = ['model_kwargs', 'epoch', 'metadata', 'optimizer_kwargs',
+            'scheduler_kwargs', 'scheduler_state_dict']
```
stephengreen, line 44: Is this everything? What about "version" (the Dingo version, which is now saved with the model)? Do we need the "scheduler_state_dict" if we are only doing inference?

stephengreen: We have to keep in mind to update this if we change the dicts saved in the `PosteriorModel` class. (One could also use a `DingoDataset` for saving, which tracks which keys to save, but that is a pretty big change from the present implementation, so I would not suggest it.)

dingo/core/utils/pt_to_hdf5.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # Note we do not save optimizer_state_dict which is
+        # not needed at inference time and saves a significant
+        # amount of disk space.
+
+        # Metadata for CVMFS LVK distribution
+        # This needs to be exactly the same as the "basename" of the hdf5 file
+        f.attrs['CANONICAL_FILE_BASENAME'] = os.path.basename(out_file_name)
+        # Add a few metadata entries as attributes
+        f.attrs['approximant'] = d['metadata']['dataset_settings']['waveform_generator']['approximant']
```
stephengreen, line 72: What is the reason for this? To just have easier access to the approximant?

dingo/core/models/posterior_model.py
stephengreen: Ah, I see. The PyTorch `state_dict` object is an `OrderedDict`.

stephengreen (changes requested):
Looks great. There are just a couple small comments to address. Could you also confirm that this is well tested?

dingo/core/models/posterior_model.py
mpuerrer: Exactly.

dingo/core/utils/pt_to_hdf5.py
mpuerrer: Yes, this is just an example. I think it is useful to have it top-level, but we might also encode it in the model filename.

dingo/core/utils/pt_to_hdf5.py
mpuerrer: You're right, I think we can drop `'optimizer_kwargs', 'scheduler_kwargs', 'scheduler_state_dict'`.
Yes, any new dicts we add in `PosteriorModel` would need to be added to this list.

mpuerrer:
One last thing to check: do older networks also have the keys `['model_kwargs', 'model_state_dict', 'epoch', 'metadata']`? If not I'll add a check.

dingo/core/utils/pt_to_hdf5.py
```diff
@@ -70,6 +69,9 @@ def main():
         f.attrs['CANONICAL_FILE_BASENAME'] = os.path.basename(out_file_name)
         # Add a few metadata entries as attributes
         f.attrs['approximant'] = d['metadata']['dataset_settings']['waveform_generator']['approximant']
+        f.attrs['epoch'] = d['epoch']
+        # Add the dingo version used for training
+        f.attrs['dingo_version'] = str(d.get('version'))
```
stephengreen, line 74: Can this key be called simply "version"? That would keep it consistent with other Dingo files.
```suggestion
        f.attrs['version'] = str(d.get('version'))
```

stephengreen (changes requested):
Thanks @mpuerrer. I think one other change that is needed is to modify `dingo_ls` to print information for HDF5 models, just like it does for PT models.

For this it would be useful to have an identifier within the HDF5 file that specifies its type. For classes that inherit from `DingoDataset` we have an attribute called `dataset_type` that does this, e.g., https://github.com/dingo-gw/dingo/blob/520dba8f0e04163c563a5e765bd6a87fe0a3ac2f/dingo/core/result.py#L55. This could also be bypassed by looking instead for the `model_version_number` key, for example.

Ultimately it might make sense to also inherit from `DingoDataset`, but probably not worth the effort now.

stephengreen:
One other suggestion would be to add the new script to the list of commands in the documentation: https://dingo-gw.readthedocs.io/en/latest/overview.html#summary-of-commands

mpuerrer:
I've added the `dataset_type`, updated `dingo_ls` to print info about the HDF5 file (pprinting the dicts 'model_kwargs' and 'metadata'), and mentioned `dingo_pt_to_hdf5` in the documentation. Should be good to go.

dingo/gw/ls_cli.py
```diff
@@ -122,6 +124,20 @@ def ls():
...
+                print(f"Dingo version: {f.attrs['version']}")
+                print("\nTrained model\n" + "================\n")
+
+                print(f"Approximant: {f.attrs['approximant']}")
+                print(f"Epoch: {f.attrs['epoch']}")
+
+                for d in ['model_kwargs', 'metadata']:
+                    json_data = json.loads(f['serialized_dicts'][d][()])
+                    print(f"\n{d}:\n" + "-"*(len(d)+1))
+                    pprint(json_data)
```
stephengreen, line 139: For consistency, I think it's better to just print exactly what we do for a PT model but it's not a huge deal and this should be equivalent.

https://github.com/dingo-gw/dingo/blob/main/dingo/gw/ls_cli.py#L21-L33

stephengreen (approved):
Thank you -- looks great!

dingo/gw/ls_cli.py
mpuerrer: The last commit makes the output conform more closely to the one printed for `.pt` files. I left the `json` data alone, though.

stephengreen (approved):

PR #171, opened 2023-03-29
Fix bug in `get_mismatch`

max-dax:
Our `get_mismatch` function `dingo.gw.gwutils` would permanently whiten the input when an ASD was provided, which would lead to downstream bugs if the waveform data is used for anything else after whitening. I fixed this, and also added support for batched data (frequency axis is as always -1).


stephengreen (approved):
Looks good, but where is this function used? It only seems to be relevant for tests.

max-dax:
We should probably use it here https://github.com/dingo-gw/dingo/blob/2d6cc239ee2b9a2eddac04bf943d236dcf54c67c/dingo/gw/SVD.py#L116.

PR #173, opened 2023-03-31
Adding more in depth docs/tutorials

nihargupte-ph:
The goal of this PR is to add 3 tutorials to `docs/source`. 

Tutorial 1: Toy Example
> This should be able to run on a local machine in < 10 minutes. The goal here is to just explain some settings about the code and what it is outputting

Tutorial 2: Normalizing Flow without GNPE
> This would contain production settings to train a normalizing flow. The goal here is to explain the production settings and how to change them to suit the user's needs. The problem with this tutorial is I have not trained a network without GNPE for a very long time, so actually, I'm not sure what the optimal settings are. For now, I was just going to put the same settings as GNPE. 

Tutorial 3: Normalizing Flow with GNPE
> Same as tutorial 2, except including the GNPE. These would be the recommended settings. 

Note it is still a draft at the moment since need to fix some grammar and add more tutorials but feel free to add things.


stephengreen:
Thanks Nihar. Regarding naming, I think it would be good to call the examples something like

* Toy NPE model
* NPE model (production)
* GNPE model (production)

That way we avoid naming things by what they are *not*, e.g., "No GNPE".

mpuerrer:
I have make some edits to the toy example tutorial, mostly minor things to make the wording more concise. Let me know if there is anything you disagree with. Will have a look at the other tutorials next.

examples/gnpe_model/GW150914.ini
```diff
@@ -0,0 +1,47 @@
+################################################################################
+##  Job submission arguments
+################################################################################
+
+local = True
+accounting = dingo
+request-cpus-importance-sampling = 64
+n-parallel = 4
```
stephengreen, line 8:
```suggestion
```

examples/gnpe_model/GW150914.ini
```diff
@@ -0,0 +1,47 @@
...
+sampling-requirements = [TARGET.CUDAGlobalMemoryMb>40000]
+# extra-lines = [+WantsGPUNode = True]
+simple-submission = false
```
stephengreen, line 11:
```suggestion
```

stephengreen (commented):
Thanks for preparing this. I looked through the examples and made a number of suggestions, mostly pertaining to simplifying the toy model. I want to know also whether you have tested these examples, particularly the toy model?

I will look next at the documentation.

docs/source/example_toy_npe_model.md
```diff
@@ -0,0 +1,127 @@
...
+The `examples/toy_npe_model/asd_dataset_settings.yaml` file includes several attributes. `f_s` is the sampling frequency in Hz, `time_psd` is the length of time used for an ASD estimate, and `T` is the duration of each ASD segment. Thus, the value of `time_psd`/`T` gives the number of segments analyzed to estimate one ASD. To avoid spectral leakage, a window is applied to each segment. We use the standard window used in LVK analyses, a Tukey window with a roll off of $\alpha=0.4$. The next attribute, `num_psds_max=1`, defines the number of ASDs stored in the ASD dataset. For now, we will use only one. See the next [tutorial](example_npe_model.md) for a more advanced setup.
+
+Step 3 Training the network
+---------------------------
+
+To train the network run
+
+```
+sed -i 's+/path/to/waveform_dataset.hdf5+'"$TRAIN_DIR"'/waveform_dataset.hdf5+g' examples/toy_npe_model/train_settings.yaml
+sed -i 's+/path/to/asd_dataset.hdf5+'"$TRAIN_DIR"'/asd_dataset_folder/asds_O1.hdf5+g' examples/toy_npe_model/train_settings.yaml
```
stephengreen, line 90: We could assume already in the YAML file the simple directory structure described above (using the `training/` directory) for model files. This way we can avoid asking the user to edit the files in this way.

docs/source/example_toy_npe_model.md
```diff
@@ -0,0 +1,127 @@
...
+The final step is to do inference, for example on GW150914. To do this we will use
+[dingo_pipe](dingo_pipe.md). For a local run execute:
+
+```
+sed -i "s|TRAIN_DIR/|$TRAIN_DIR/|g" examples/toy_npe_model/GW150914_toy.ini
+sed -i "s|/path/to/model.pt|$TRAIN_DIR/model_latest.pt|g" examples/toy_npe_model/GW150914_toy.ini
+dingo_pipe examples/toy_npe_model/GW150914.ini
+```
+
+This will generate files which are described in [dingo_pipe](dingo_pipe.md). To see the results, take a look in `$TRAIN_DIR/outdir_GW150914`.
```
stephengreen, line 116: Don't put the output of inference in the training directory.

```suggestion
This will generate files which are described in [dingo_pipe](dingo_pipe.md). To see the results, take a look in `outdir_GW150914`.
```

docs/source/example_toy_npe_model.md
```diff
@@ -0,0 +1,127 @@
...
+
+We can load and manipulate the data with the following code. For example, here we create a cornerplot
```
stephengreen, line 118: It's okay to show how to make a plot, but dingo_pipe will already produce this corner plot, and that should be mentioned.

docs/source/example_npe_model.md
```diff
@@ -0,0 +1,127 @@
...
+To generate an ASD dataset we can run the same command as in the previous tutorial.
+
+```
+dingo_generate_asd_dataset --settings_file examples/npe_model/asd_dataset_settings_fiducial.yaml --data_dir $TRAIN_DIR/asd_dataset_folder -out_name $TRAIN_DIR/asds_O1_fiducial.hdf5
+```
+
+However, this time, during training we will need two sets of ASDs. The first one will be
+fixed during the initial training -- this is the fiducial dataset generated above.
+This dataset will contain only  a single ASD. The second ASDDataset will contain many
+ASDs and is used during the fine tuning stage. To generate this second dataset run
```
stephengreen, line 64: This will seem somewhat mysterious to a reader. Why are we using two sets of ASDs?

docs/source/example_npe_model.md
```diff
@@ -0,0 +1,127 @@
...
+
+Also notice there are now two training stages `stage_0` and `stage_1`. In `stage_0` a fixed ASD is used and the reduced basis layer
+is frozen. Then in `stage_1` all ASDs are used and the reduced basis layer is unfrozen. 
+
+The main difference in the local settings is that the `device` is set to `CUDA`.
+Note if you have multiple GPUs on the machine, you can select which GPU to use
+by running 
+
+```{important}
+It is recommended to have at least 40 GB of GPU memory on the device.
```
stephengreen, line 109: What should a user do if they have < 40 GB of GPU RAM? (Reduce the batch size, or network size.)

docs/source/example_gnpe_model.md
```diff
@@ -0,0 +1,90 @@
...
+dingo_generate_asd_dataset --settings_file $(pwd)/examples/gnpe_model/asd_dataset_settings.yaml --data_dir
+$TRAIN_DIR/asd_dataset_folder -out_name $TRAIN_DIR/asds_O1.hdf5
+```
+
+
+Step 3 Training the network
+---------------------------
+
+Now we are ready for training using GNPE. Here we need to train two networks, one which estimates the time of arrival 
+in the detectors and one which does the full inference task. First we train the initialization network for the detector times with:
```
stephengreen, line 51: There needs to be some discussion of *why* we are doing what we are doing, e.g., what GNPE is, why it needs two networks, why the initialization network is smaller, etc. It may also be helpful to visualize by including the flowchart from the quickstart tutorial.

docs/source/example_toy_npe_model.md
```diff
@@ -0,0 +1,127 @@
...
+cd dingo
+mkdir $(pwd)/toy_npe_model_train_dir
+export TRAIN_DIR=$(pwd)/toy_npe_model_train_dir
+```
+
+To generate a waveform dataset run
+
+```
+dingo_generate_dataset --settings examples/toy_npe_model/waveform_dataset_settings.yaml --out_file $TRAIN_DIR/waveform_dataset.hdf5
+```
```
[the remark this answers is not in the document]
nihargupte-ph: That is true, I think there should be the download from the remote repo instead of the devel install. Though I think setting the $TRAIN_DIR may be convenient, since that way the user can just run the later commands without having to really think about it too much. But yeah having the sed lines later may overcomplicate things. I'll see if I can reorganize things to make it a bit easier. Maybe the file structure could be something like:

toy_npe_model/
   train_dir/
      waveform_dataset.hdf5
      asd_dataset/
         asd_dataset.hdf5
         tmp/
     training/
        train_settings.yaml
        local_settings.yaml
        (SVDs optionally)
        model_latest.pt
   inference/
      GW150914.ini
      outdir_GW150914/
          dingo_pipe_stuff

examples/toy_npe_model/GW150914.ini
```diff
@@ -0,0 +1,44 @@
...
+
+################################################################################
+##  Sampler arguments
+################################################################################
+
+model = /path/to/model.pt
+device = 'cpu'
+num-gnpe-iterations = 30
+num-samples = 50000
+batch-size = 50000
```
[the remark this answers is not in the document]
nihargupte-ph: > Thanks for preparing this. I looked through the examples and made a number of suggestions, mostly pertaining to simplifying the toy model. I want to know also whether you have tested these examples, particularly the toy model?
> 
> I will look next at the documentation.

I've ran the toy example locally, but the problem I'm having is that with the settings I'm using the posterior comes out really bad. This is expected, but the problem is when plotting the posterior there is an error because of the importance weights.

examples/npe_model/waveform_dataset_settings.yaml
```diff
@@ -0,0 +1,40 @@
...
+  chirp_mass: bilby.gw.prior.UniformInComponentsChirpMass(minimum=15.0, maximum=150.0)
+  mass_ratio: bilby.gw.prior.UniformInComponentsMassRatio(minimum=0.125, maximum=1.0)
+  phase: default
+  a_1: bilby.core.prior.Uniform(minimum=0.0, maximum=0.99)
+  a_2: bilby.core.prior.Uniform(minimum=0.0, maximum=0.99)
+  tilt_1: default
+  tilt_2: default
+  phi_12: default
+  phi_jl: default
+  theta_jn: default
```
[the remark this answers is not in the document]
nihargupte-ph: I can update this, though which aligned spin model? I think SEOBNRv4 makes sense since it's the simplest aligned spin model that I've tested. Though I haven't trained a pure NPE network.

examples/gnpe_model/train_settings_init.yaml
```diff
@@ -95,4 +94,4 @@ local:
     num_cpus: 16
     memory_cpus: 128000
     num_gpus: 1
-    memory_gpus: 8000
+    memory_gpus: 8000
```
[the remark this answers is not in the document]
nihargupte-ph: Hmm, might just be something local to the MPI-IS cluster.

docs/source/example_npe_model.md
nihargupte-ph: https://github.com/dingo-gw/dingo/pull/173/commits/1963348ee56b5af159d1f31d2ddab2dc7652faa0

docs/source/example_npe_model.md
nihargupte-ph: https://github.com/dingo-gw/dingo/pull/173/commits/fad0ea2cd785ec8978b6aba2f5703c83e635eb8c

docs/source/example_toy_npe_model.md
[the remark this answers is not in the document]
stephengreen: You could flatten it further. Put all the config files from the relevant examples directory at the lowest level. Then put all the training data files in a subdirectory, the trained models in another, and the inference output in the final one:

```
toy_npe_model/

    #  config files
    waveform_dataset_settings.yaml
    asd_dataset_settings.yaml
    train_settings.yaml
    GW150914.ini

    training_data/
        waveform_dataset.hdf5
        asd_dataset.hdf5
        tmp/    #  Contains temporary files from ASD dataset generation

    models/
        model_050.pt
        model_stage_0.pt
        model_latest.pt
        history.txt
        #  etc...

    outdir_GW150914/
        #  dingo_pipe output
```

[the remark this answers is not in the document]
stephengreen: In fact, it might be a good idea to explicitly list that directory structure (as written above) in the documentation for the toy  example. That would be a way to describe all the various files and how they relate to each other.

examples/toy_npe_model/train_settings.yaml
```diff
@@ -1,5 +1,5 @@
 data:
-  waveform_dataset_path: /path/to/waveform_dataset.hdf5  # Contains intrinsic waveforms
+  waveform_dataset_path: /home/n/Documents/Research/Dingo/dingo/toy_example_train_dir/waveform_dataset.hdf5  # Contains intrinsic waveforms
   train_fraction: 0.95
   window:  # Needed to calculate window factor for simulated data
```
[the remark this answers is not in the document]
nihargupte-ph: Note sure if I understood this change, it should be /path/to/waveform_dataset.hdf5 right?

examples/npe_model/train_settings.yaml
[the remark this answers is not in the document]
nihargupte-ph: Ok, but what about in the the toy example? I've set it to sample the phase for now since that is an advanced feature?

examples/toy_npe_model/GW150914.ini
stephengreen, line 1: The current setup will likely lead to errors, since it will attempt a synthetic phase reconstruction, even though the phase is already present.  To fix this, either:
* Drop the `phase` fro the inference parameters specified in the train settings file; or
* Set `importance-sampling-settings = {}` in the INI file.

examples/npe_model/train_settings.yaml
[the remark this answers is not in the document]
stephengreen: I would use the synthetic phase in the production example, but not in the toy example. That means that in production, drop phase from the inference parameters. And in the toy example, leave it in but make sure dingo_pipe doesn't try to sample synthetic phase.

examples/npe_model/train_settings.yaml
```diff
@@ -20,7 +20,18 @@ data:
...
+  - mass_ratio
+  - chi_1
+  - chi_2
+  - theta_jn
+  - dec
+  - ra
+  - geocent_time
+  - luminosity_distance
+  - psi
+  - phase
```
stephengreen, line 34:
```suggestion
```

examples/npe_model/train_settings.yaml
```diff
@@ -67,7 +78,7 @@ training:
 
   stage_1:
     epochs: 150
-    asd_dataset_path: /path/to/asds.hdf5 # this should contain many different ASDS per detector for finetuning
+    asd_dataset_path: training_data/asd_dataset_folder/asds_O1.hdf5 # this should contain many different ASDS per detector for finetuning
```
stephengreen, line 81: Generally it is better not to include "folder" within the folder name.

docs/source/example_gnpe_model.md
```diff
@@ -0,0 +1,131 @@
...
+
+First generate the directory structure:
+
+```
+cd gnpe_model
+mkdir training_data
+mkdir training
+mkdir training/main_train_dir
+mkdir training/init_train_dir
+
```
stephengreen, line 59: This file isn't rendering properly. You need to close this code block.

stephengreen (changes requested):
Thanks for making all these updates. It's looking much better. There are a few new comments in the code that need to be addressed.

stephengreen:
> It also needs to be decided how to integrate these examples together with the "quickstart tutorial", which has significant overlap. Maybe, e.g., the quickstart could be replaced by the toy example, but the flow chart could go in the GNPE example. A simplified flow chart could also be used for the two NPE examples.

This should also be addressed at some point, but I understand if there isn't enough time now.

examples/npe_model/train_settings.yaml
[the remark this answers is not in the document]
nihargupte-ph: Ok, I think this should be working then because I ran through the toy example and was able to generate samples when inferring the phase.

examples/toy_npe_model/GW150914.ini
nihargupte-ph: I think in the toy example not passing the `importance-sampling-settings` just makes it so it doesn't sample the synthetic phase, at least I'm able to produce a result with a phase column without it

examples/gnpe_model/asd_dataset_settings.yaml
```diff
@@ -1,13 +1,22 @@
...
+#  f_max: 2048      # defaults to f_s/2
   f_s: 4096
   time_psd: 1024
   T: 8
   window:
     roll_off: 0.4
     type: tukey
   time_gap: 0          # specifies the time skipped between to consecutive PSD estimates. If set < 0, the time segments overlap
-  num_psds_max: 0  # if set > 0, only a subset of all available PSDs will be used
+  num_psds_max: 20   # if set > 0, only a subset of all available PSDs will be used
```
stephengreen, line 11: This is a small-ish number. Should you maybe comment on it?

examples/gnpe_model/asd_dataset_settings.yaml
```diff
@@ -1,13 +1,22 @@
...
+#  channels:
+#    H1: H1:DCS-CALIB_STRAIN_C02
+#    L1: L1:DCS-CALIB_STRAIN_C02
   detectors:
     - H1
-    - L1
```
stephengreen: Only one detector?

stephengreen (changes requested):
Looking really good, I just added a couple comments to address.

Could you please also update the toy example to not use importance sampling? That would simplify it more. Set `importance-sample = False` in the INI file (and update docs).

Thanks.

stephengreen (approved):

PR #174, opened 2023-03-31
Wandb import

jonaswildberger:
Moved all calls to the `wandb` library into `try`, `except` statements, therefore dropping the requirement to have it installed. Some further improvements were made to use it in a more generic way


stephengreen (changes requested):
Looks good. The only remaining task is to change the pip requirements.

stephengreen:
> Looks good. The only remaining task is to change the pip requirements.

Done

stephengreen (approved):

PR #178, opened 2023-04-10
Gwsignal interface

hectorestelles:
This MR introduces:

- Adds new class GWSignalWaveformGenerator for interfacing GWSignal.
- Allows choosing the generator from the dataset settings.

To be more tested: calling LAL models from the GWSignalWaveformGenerator, but this is not crucial for O4a since LAL models will still be callable through the LALSim interfaces.


stephengreen (commented):
A couple comments / questions:

* The name GWSignal should be replaced by NewInterface throughout so as to avoid confusion the Dingo GWSignal class.
* There looks to be a lot of duplicated code in the new WaveformGenerator compared with the old one. Could they be restructed to both inherit from a common superclass? I realize that that is a lot of work, so for now maybe just move the new WaveformGenerator to a new module, since the current file is quite long.

dingo/gw/dataset/generate_dataset.py
```diff
@@ -149,10 +149,18 @@ def generate_dataset(settings: Dict, num_processes: int) -> WaveformDataset:
 
     prior = build_prior_with_defaults(settings["intrinsic_prior"])
     domain = build_domain(settings["domain"])
-    waveform_generator = WaveformGenerator(
-        domain=domain,
-        **settings["waveform_generator"],
-    )
+
+    gwsignalflag = settings.get("gwsignal_generator",False)
```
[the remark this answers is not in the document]
stephengreen: Also the name of the associated flag in the config file should be updated. In fact, it probably makes sense to place the flag under `waveform_generator` in the config file. That way we don't have to pass around this extra waveform generator argument everywhere.

dingo/gw/result.py
```diff
@@ -308,6 +309,21 @@ def _build_likelihood(
...
 
+        # The detector reference positions during likelihood evaluation should be based
+        # on the event time, since any post-correction to account for the training
+        # reference time has already been applied to the samples.
+
+        if self.event_metadata is not None and "time_event" in self.event_metadata:
+            t_ref = self.event_metadata["time_event"]
+        else:
+            t_ref = self.base_metadata["train_settings"]["data"]["ref_time"]
+
```
stephengreen, line 320: I don't believe this code belongs here. It was moved to a property `t_ref`.
```suggestion
```

dingo/gw/result.py
```diff
@@ -308,6 +309,21 @@ def _build_likelihood(
...
+            t_ref = self.event_metadata["time_event"]
+        else:
+            t_ref = self.base_metadata["train_settings"]["data"]["ref_time"]
+        
+        #Select waveform generator, LAL or GWSignal
+        if self.base_metadata["dataset_settings"].get("gwsignal_generator") is None:
+            waveform_generator = WaveformGenerator
+        else:
+            waveform_generator = NewInterfaceWaveformGenerator
+
```
stephengreen, line 326: Again, this is not needed if you put the choice of waveform generator within `wfg_kwargs` from the start.

dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -3,12 +3,15 @@
...
 import numpy as np
+import astropy.units as u
 from typing import Dict, List, Tuple, Union
 from numbers import Number
 import warnings
 import lal
 import lalsimulation as LS
 import pandas as pd
+from lalsimulation.gwsignal.core import waveform as gws_wfm
+from lalsimulation.gwsignal.models import gwsignal_get_waveform_generator
```
stephengreen, line 14: Rename these to refer to new interface, not gwsignal.

dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -763,6 +766,603 @@ def generate_TD_waveform(self, parameters_lal: Tuple) -> Dict[str, np.ndarray]:
...
+            {m: {"h_plus": hp_m, "h_cross": hc_m} for m in [-l_max,...,0,...,l_max]}
+            Each contribution h_m transforms as exp(-1j * m * phase) under phase shifts
+            (for fixed self.spin_conversion_phase, see above).
+        """
+        if not isinstance(parameters, dict):
+            raise ValueError("parameters should be a dictionary, but got", parameters)
+        elif not isinstance(list(parameters.values())[0], float):
+            raise ValueError("parameters dictionary must contain floats", parameters)
+
+        generator = gwsignal_get_waveform_generator(self.approximant_str)
```
stephengreen, line 1196: What is the runtime for this function? Does it make sense to save the generator as an attribute to be able to access it faster?

dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -763,6 +766,603 @@ def generate_TD_waveform(self, parameters_lal: Tuple) -> Dict[str, np.ndarray]:
         pol_dict = {"h_plus": h_plus, "h_cross": h_cross}
         return pol_dict
 
+class NewInterfaceWaveformGenerator:
```
stephengreen, line 769: In general this class is not easy to review since it duplicates much of the code from `WaveformGenerator`. Could it be made to subclass `WaveformGenerator` and only replace certain methods?

hectorestelles: Hi Stephen, I committed now a change to make NewInterfaceWaveformGenerator a subclass of WaveformGenerator. This has reduced the amount of duplicated code. More duplication can be avoided in the method generate_hplus_hcross_m, which I think can be defined only once, but this requires some extra modifications, I will work on it  (and on the rest of the comments above) tomorrow.

dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -85,14 +108,12 @@ def __init__(
...
 
-        self.lal_params = None
-        if mode_list is not None:
-            self.lal_params = self.setup_mode_array(mode_list)
-
         self.transform = transform
         self._spin_conversion_phase = None
         self.spin_conversion_phase = spin_conversion_phase
 
+        self.new_interface = new_interface
```
stephengreen, line 115: It is slightly awkward for the superclass `WaveformGenerator` to have to know about the new interface. I see that you do this so that you can know whether to call `_convert_parameters_to_lal_frame` or `_convert_paramters_new_interface`. However, this would lead to errors if we instantiate `WaveformGenerator` with `new_interface = True`.

These issues could be avoided by renaming `_convert_parameters_to_lal_frame` to `_convert_parameters`, and overriding this method in the subclass. That way the desired method will get called, depending on the calling class type.

dingo/gw/waveform_generator/waveform_generator.py
stephengreen: Thanks. Yeah I see there is still some room for reducing duplication. If you don't have time to do it now, let me know.

stephengreen (changes requested):
Thanks for the updates. I added a few more comments on the inheritance.

dingo/gw/waveform_generator/waveform_generator.py
hectorestelles: It is quite fast, just calls the generator class of the approximant. Storing the approximant generator as an attribute produces a pickle error I couldn't manage to solve, therefore I think it is fine leaving calling this function when needed

dingo/gw/waveform_generator/waveform_generator.py
hectorestelles: In a next version, reducing more core needs some extra thinking on how to manage certain things, so better to spend more time, for a next version

dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -72,7 +82,13 @@ def __init__(
...
         else:
             self.approximant_str = approximant
-            self.approximant = LS.GetApproximantFromString(approximant)
+            self.lal_params = None
+            try:
+                self.approximant = LS.GetApproximantFromString(approximant)
+                if mode_list is not None:
+                    self.lal_params = self.setup_mode_array(mode_list)
+            except:
+                pass
```
stephengreen, line 91: Why was this moved from where it was into the `else` and `try`/`except`?

dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -3,24 +3,35 @@
...
+
+try:
+    from lalsimulation.gwsignal.core import waveform as gws_wfm
+    from lalsimulation.gwsignal.models import (
+        gwsignal_get_waveform_generator as new_interface_get_waveform_generator,
+    )
+except:
+    print(
+        "LALsimulation does not contain gwsignal module, you will not be able to employ new interface."
+    )
```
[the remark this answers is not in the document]
hectorestelles: Error message deleted, but imports couldn't be included inside init method, otherwise they are not accessible outside, and storing the imported methods as attributes produces a pickle error in the code. I've tested and with an environment that does not contain latest lalsimulation, it will just not import that, without errors

stephengreen (approved):
Thank you, I just had one question that I wanted to know the answer to.

dingo/gw/waveform_generator/waveform_generator.py
hectorestelles: Yes, argument and attribute eliminated

dingo/gw/waveform_generator/waveform_generator.py
hectorestelles: With non-LAL approximants, LS.GetApproximantFromString produces an error.

hectorestelles: And now SEOBNRv5PHM will pass through this code, since it initializes WaveformGenerator, so I though it was safer this way

PR #179, opened 2023-04-12
Pepointed dingo.gw.pipe.dingo_result to dingo.pipe.dingo_result

nihargupte-ph:
Small bug when using dingo pipe to merge results.


stephengreen (approved):
Thanks for catching this.

PR #180, opened 2023-04-18
Corrected for calibration marginalized likelihood

nihargupte-ph:
There was a slight mistake in calculating the calibration marginalized likelihood which showed up in a difference between the offiical bilby samples and dingo's samples. This has now been corrected, the reason for this bug is included in the following calculation: 

image


stephengreen (approved):
Looks good! Great job catching this subtle error.

PR #181, opened 2023-04-19
Rename the imported logger to "dingo_pipe"

mpuerrer:
The change made here is a bit lazy, but simply replacing the (bilby_pipe) `logger` with a new one in dingo_pipe code would lead to inconsistent output because we are calling bilby_pipe code in many places and that is using it's own logger. So, to leave the bilby_pipe code unchanged we take their logger and simply rename it.


stephengreen (approved):
Perfect, thank you!

PR #184, opened 2023-04-20
Add ability to turn off importance sampling, create summary pages

stephengreen:
This makes two additions to dingo_pipe:

* `importance-sample` option (default True). By setting this to False, we can turn off importance sampling.
* `create-summary`. This is taken over from bilby_pipe. It calls PESummary at the end of the run.

### DIFF
--- a/dingo/pipe/dag_creator.py
+++ b/dingo/pipe/dag_creator.py
@@ -10,6 +10,7 @@ from bilby_pipe.utils import BilbyPipeError, logger
 from dingo.pipe.nodes.generation_node import GenerationNode
 from .nodes.importance_sampling_node import ImportanceSamplingNode
 from .nodes.merge_node import MergeNode
+from .nodes.pe_summary_node import PESummaryNode
 from .nodes.plot_node import PlotNode
 from .nodes.sampling_node import SamplingNode
 
@@ -42,7 +43,7 @@ def get_parallel_list(inputs):
         return [f"part{idx}" for idx in range(inputs.n_parallel)]
 
 
-def generate_dag(inputs):
+def generate_dag(inputs, model_args):
     inputs = copy.deepcopy(inputs)
     dag = Dag(inputs)
     trigger_times = get_trigger_time_list(inputs)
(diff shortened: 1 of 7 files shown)


dingo/pipe/dag_creator.py
```diff
@@ -78,63 +79,67 @@ def generate_dag(inputs):
...
+                #       (Automatically calculates evidence.)
+                merge_node = MergeNode(
+                    inputs=inputs,
+                    parallel_node_list=parallel_node_list,
+                    dag=dag,
+                )
+                merged_importance_sampling_node_list.append(merge_node)
+
+    else:
+        merged_importance_sampling_node_list = sampling_node_list
```
mpuerrer, line 142: Consider renaming `merged_importance_sampling_node_list` to `merged_sampling_node_list` or `merged_node_list`.

dingo/pipe/nodes/sampling_node.py
```diff
@@ -65,3 +65,7 @@ def samples_file(self):
         return os.path.join(
             self.inputs.result_directory, self.label + ".hdf5"
         )
+
+    @property
+    def result_file(self):
+        return self.samples_file
```
mpuerrer, line 71: Given that `self.samples_file` is used should the TODO in line 64 be removed?

mpuerrer (commented):
I'm not very familiar with the pipe aspect of the code, but AFAIK it looks sane! Perhaps also mention that summarypages is currently a stub.

dingo/pipe/nodes/sampling_node.py
stephengreen: Ah, I have to check whether that's referred to. We need `result_file` for `PESummary`. But I'm not sure whether the other one is needed. I'll fix later.

dingo/pipe/dag_creator.py
stephengreen: I agree the naming is brutal; I will fix in a later update since I do not want to risk making an error.

PR #185, opened 2023-04-20
Automatically convert to Bilby spins when outputting to PESummary

stephengreen:
Also add a new method `get_samples_bilby_phase` to GW Result class.

### DIFF
--- a/dingo/gw/result.py
+++ b/dingo/gw/result.py
@@ -506,6 +506,25 @@ class Result(CoreResult):
 
         print(f"Done. This took {time.time() - t0:.2f} s.")
 
+    def get_samples_bilby_phase(self):
+        """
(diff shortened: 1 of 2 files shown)


dingo/gw/result.py
```diff
@@ -506,6 +506,25 @@ def sample_synthetic_phase(
 
         print(f"Done. This took {time.time() - t0:.2f} s.")
 
+    def get_samples_bilby_phase(self):
```
mpuerrer, line 509: I don't see `get_samples_bilby_phase()` getting used anywhere?

stephengreen: It is not called internally. It's for the user if they want to convert.

mpuerrer (approved):

PR #189, opened 2023-04-21
updated T and asd from O2 to O1

nihargupte-ph:


examples/toy_npe_model/asd_dataset_settings.yaml
```diff
@@ -3,7 +3,7 @@ dataset_settings:
 #  f_max: 2048      # defaults to f_s/2
   f_s: 4096
   time_psd: 1024
-  T: 8
+  T: 4
```
stephengreen, line 6: We're using 4 s PSDs here; is that consistent with the rest of the "toy example" settings?

examples/toy_npe_model/training/local_settings.yaml
stephengreen, line 1: Don't save the local settings.

examples/toy_npe_model/training/model_latest.pt
stephengreen, line 1: Don't save networks.

examples/toy_npe_model/training_data/asd_dataset/tmp/O1/psd_time_segments.pkl
stephengreen, line 1: Remove data files.

stephengreen (changes requested):
Thanks. I think you accidentally committed data files. I also want to confirm that 4s is being used throughout the toy example.

examples/toy_npe_model/asd_dataset_settings.yaml
nihargupte-ph: Right, it was changed to 8 in the last commit which broke the example. So I changed it back to 4 to get things to work.

examples/toy_npe_model/training/model_latest.pt
nihargupte-ph: oops, I rewrote the history to fix this....

stephengreen (changes requested):
I think the train settings file (and associated docs) for the toy model also has to be updated to 4s.

stephengreen (approved):

PR #192, opened 2023-04-25
hdf5 locking fix

nihargupte-ph:
For some reason a fresh install of dingo and executing dingo_pipe on a .ini file gives an error stating that a BilbyMainInput doesn't have a `disable_hdf5_locking` attribute. I'm not actually sure if this should be set from the command line arguments via args.disable_hdf5_locking, but setting it to None fixes the issue.

### DIFF
--- a/dingo/pipe/nodes/importance_sampling_node.py
+++ b/dingo/pipe/nodes/importance_sampling_node.py
@@ -51,8 +51,8 @@ class ImportanceSamplingNode(AnalysisNode):
 
         self.extra_lines.extend(self._checkpoint_submit_lines())
         env_vars = []
-        if self.request_cpus > 1:
-            env_vars.append("OMP_NUM_THREADS=1")
+        # if self.request_cpus > 1:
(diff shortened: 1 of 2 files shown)


mpuerrer:
Could this be a problem that is specific to AEI clusters? Do you get the same issue on your laptop or on CIT?

stephengreen:
I have have implemented changes corresponding to

https://git.ligo.org/lscsoft/bilby_pipe/-/merge_requests/529
https://git.ligo.org/lscsoft/bilby_pipe/-/merge_requests/534

`dingo_pipe` requires `bilby_pipe >= 1.0.9` to work now, so this is a requirement in `pyproject.toml`.

Please check that `dingo_pipe` works, since I have not tested these changes.

nihargupte-ph:
> Please check that `dingo_pipe` works, since I have not tested these changes.

Works on my local machine

stephengreen (approved):
I reviewed this, but I also wrote it, so maybe somebody else should have a look too.

PR #194, opened 2023-05-08
Add script to make pp-plots with optional confidence bands

mpuerrer:
A reworking of Jonas' code to make a p-p plot with added binomial confidence bands.

I noticed that `weighted_percentile_of_score()` isn't actually used in this script, but it must have been used earlier when computing the percentile data, so it could be moved into another script, or this script could be extended to include code used for computing the percentile data.

### DIFF
--- a/dingo/gw/inference/make_pp_plot.py
+++ b/dingo/gw/inference/make_pp_plot.py
@@ -1,0 +1,173 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy import stats
+import pandas as pd
+import argparse
+import textwrap
+
+
+def weighted_percentile_of_score(samples, value, weights=None):
+    """
+    Compute the percentile rank of a score relative to a list of scores
+    with attached weights.
+    
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


dingo/gw/inference/make_pp_plot.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # Calculate the cumulative sum of the weights and normalize to get the CDF
+    cdf = np.cumsum(sorted_weights) / sorted_weights.sum()
+
+    # Find the index of the first element in the sorted samples array
+    # that is greater than or equal to the given value
+    index = np.searchsorted(sorted_samples, value)
+
+    # Look up the corresponding CDF value at that index to get the percentile
+    percentile = cdf[index] * 100
```
stephengreen, line 46: For `value` lying between two samples, this gives the percentile of the larger sample. There are other possibilities, e.g., given as options to the `method` argument of [numpy.percentile()](https://numpy.org/doc/stable/reference/generated/numpy.percentile.html). Are we doing the same as Bilby?

PR #195, opened 2023-05-09
Restore compatibility for bilby_pipe < 1.0.9

stephengreen:
For compatibility with `bilby_pipe >= 1.0.9`, we assumed an attribute of the `Input` class called `disable_hdf5_locking` in #192. However, this introduced an incompatibility with older version of `bilby_pipe`. This new PR simply gives a default attribute value of `None` by accessing via `getattr()`.

I am currently testing with `bilby_pipe = 1.0.7` and `bilby_pipe = 1.1.0`.


stephengreen:
Tested, works fine.

PR #196, opened 2023-05-09
Add lal import in `dingo/gw/__init__.py`

stephengreen:
This is intended to try to resolve an incompatibility between lal / MKL / pytorch, see https://git.ligo.org/lscsoft/lalsuite/-/issues/300#note_680778.


mpuerrer:
LGTM

mpuerrer (approved):

Issue #201, opened 2023-05-25
About running the toy model in document


stephengreen:
### Discussed in https://github.com/dingo-gw/dingo/discussions/200

<div type='discussions-op-text'>

<sup>Originally posted by **thisnameavailable** May 25, 2023</sup>
I just installed dingo and want to try to use it following the toy model. When it comes to the training progress ,it raises ValueError ("Cannot update delta_f) shown in domains.py, line 134, update. I have no idea why it happens.</div>

We should update the duration of the ASD segments for the toy example from 8s to 4s.

stephengreen:
Could we also check all of the examples to make sure that the segment lengths are consistent?

PR #204, opened 2023-05-26
Quick Fix to Examples with ASD length

nihargupte-ph:
From https://github.com/dingo-gw/dingo/discussions/200 and https://github.com/dingo-gw/dingo/issues/201. Ensuring that the ASD lengths are the same


stephengreen (approved):
Looks good -- thank you!

PR #207, opened 2023-06-22
Getting Dingo Pipe To Handle Injections

nihargupte-ph:
It is useful to analyze injections with Dingo, especially because it is so fast. However, this has not yet been implemented in dingo_pipe. The goal of this PR is to accomplish this. 

:innocent: Implemented Features:
> Allows specification of `injection-parameters` in the `settings.ini` file.
> Allows one to change the waveform approximant which is being injected from the one used for training the network 
> Allows specification of PSD from trigger-time
> Allows specification of random seed for noise realization

:smiling_face_with_tear: TODO:
> Implement Zero-Noise option
> Check that the `asd-dataset` and `trigger-time` can't both be specified at the same time. As only one should be used to specify the PSD

For this^: I see we can specify multiple trigger times via: https://github.com/dingo-gw/dingo/blob/main/dingo/pipe/dag_creator.py#L49 which will generate multiple dags for data generation and sampling. I am wondering if this could be a simple way to iterate through multiple different noise realizations and at the end we can merge the result but I'm not quite sure what appending things to the generation_node_list does if it's different data.

### DIFF
--- a/dingo/pipe/sampling.py
+++ b/dingo/pipe/sampling.py
@@ -119,8 +119,9 @@ class SamplingInput(Input):
         self._load_sampler()
 
     def _load_event(self):
-        self.context = EventDataset(file_name=self.event_data_file)
-        self.event_metadata = self.context.settings
+        event_dataset = EventDataset(file_name=self.event_data_file)
+        self.context = event_dataset.data
+        self.event_metadata = event_dataset.settings
 
     def _load_sampler(self):
         """Load the sampler and set its context based on event data."""


dingo/core/result.py
nihargupte-ph, line 1: consider resetting this file

mpuerrer (changes requested):
Issue with processing `event_data_files`.

dingo/pipe/importance_sampling.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, args, unknown_args):
 
         # Samples to run on
         self.proposal_samples_file = args.proposal_samples_file
-        self.event_data_file = args.event_data_file
+        self.event_data_file = args.event_data_files[0]
```
mpuerrer, line 40: Again, `event_data_files` looks like `file1.hdf5 file2.hdf5 ...` for me. Here we operate on a single file, so 

```
self.event_data_file = args.event_data_files[0].split()[0]
```

should fix it.

mpuerrer (dismissed):

dingo/pipe/importance_sampling.py
mpuerrer: I pushed a commit to fix this and the similar issue above.

dingo/core/result.py
```diff
@@ -276,17 +276,26 @@ def importance_sample(self, num_processes: int = 1, **likelihood_kwargs):
...
+        if "log_likelihood" not in self.samples.columns:
+            print(f"Calculating {len(theta)} likelihoods.")
+            t0 = time.time()
+            log_likelihood = self.likelihood.log_likelihood_multi(
+                theta, num_processes=num_processes
+            )
+            print(f"Done. This took {time.time() - t0:.2f} seconds.")
+        else:
+            log_likelihood = self.samples["log_likelihood"]
+            log_likelihood = log_likelihood[valid_samples]
```
stephengreen, line 288: Why is it necessary to add this `else` statement? Importance sampling should be calculating the likelihood itself.

dingo/core/result.py
```diff
@@ -276,17 +276,26 @@ def importance_sample(self, num_processes: int = 1, **likelihood_kwargs):
...
 
         self.log_noise_evidence = self.likelihood.log_Zn
         self.samples["log_prior"] = log_prior
         self.samples.loc[valid_samples, "log_likelihood"] = log_likelihood
+
+        # Drop None samples where waveform failed to generate
+        self.samples = self.samples.dropna(subset=['log_likelihood'])
```
stephengreen, line 295: Are we getting NaNs in the likelihood and why is that happening?

dingo/core/result.py
```diff
@@ -547,20 +556,22 @@ def merge(cls, parts):
...
             samples_parts.append(part_dict.pop("samples"))
 
             # Make sure we are not merging incompatible results. We deleted the
             # log_evidence since this can differ among the sub-results. Note that this
             # will also raise an error if files were created with different versions of
             # dingo.
-            if not recursive_check_dicts_are_equal(part_dict, dataset_dict):
-                raise ValueError("Results to be merged must have same metadata.")
+            # if not recursive_check_dicts_are_equal(part_dict, dataset_dict):
+                # raise ValueError("Results to be merged must have same metadata.")
```
stephengreen, line 574: This was to make sure we are not merging incompatible results. Why was it necessary to comment this?

dingo/gw/waveform_generator/waveform_generator.py
stephengreen, line 1: This file does not seem to properly include @hectorestelles fix of the v5 tapering.

dingo/core/result.py
nihargupte-ph: This along with other changes to result.py I think don't belong in this PR. I've reset the file to how it is on main.

dingo/core/density/unconditional_density_estimation.py
```diff
@@ -76,7 +76,7 @@ def train_unconditional_density_estimator(
     # TODO: Allow for other types of density estimators (e.g., flow matching).
     model = NormalizingFlowPosteriorModel(
         metadata={"train_settings": settings, "base": copy.deepcopy(result.metadata)},
-        device=settings["training"]["device"],
+        device=settings["training"].get("device", "cuda"),
```
annalena-k, line 79: Might be helpful to use the fallback option that Vincent included here.

dingo/core/nn/nsf.py
```diff
@@ -207,27 +207,27 @@ def __init__(self, flow: flows.base.Flow, embedding_net: nn.Module = None):
         self.flow = flow
 
     def log_prob(self, y, *x):
-        if self.embedding_net is not None:
-            x = self.embedding_net(*x)
         if len(x) > 0:
+            if self.embedding_net is not None:
```
annalena-k, line 211: I think these changes are already in the current main branch. This should be resolved by merging main into this branch?

dingo/gw/injection.py
```diff
@@ -365,6 +365,8 @@ def injection(self, theta):
             print("self.whiten was set to True. Resetting to False.")
             self.whiten = False
 
+        if seed is not None:
+            np.random.seed(seed)
```
annalena-k, line 369: Is it enough to only set the numpy random seed or do you also need to set the torch and Bilby random seeds?

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
 
             self.create_data(args)
 
-    def save_hdf5(self):
-        """
-        Save frequency-domain strain and ASDs as DingoDataset HDF5 format.
+    def generate_injection(self, args):
+        """Generate injection consistent with trained dingo model"""
+        # loading posterior model for which we want to generate injections
+        pm = NormalizingFlowPosteriorModel(model_filename=args.model, device="cpu")
```
annalena-k, line 200: Would be better to use 
```
pm = build_model(...)
```
such that this also works for flow matching.

dingo/pipe/data_generation.py
```diff
@@ -216,21 +388,81 @@ def save_hdf5(self):
...
-            "channel_dict",
-            "data_dict",
-        ]:
-            try:
-                v = getattr(self, k)
-            except AttributeError:
-                continue
-            if v is not None:
-                settings[k] = v
+        domain = FrequencyDomain(
```
annalena-k, line 391: Would probably be good to merge this into main before the multibanding gets merged, otherwise you will have to make it compatible ...

annalena-k (commented):
Some small comments. Once main is merged into this branch, I can test it locally.

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
...
+        rho_opt = np.sqrt(sum(kappa2_list))
+        rho_opt_ifos = np.sqrt(kappa2_list)
+
+        return rho_opt_ifos, rho_opt
 
-            strain = ifo.strain_data.frequency_domain_strain
+    def create_data(self, args):
+        super().create_data(args)
+
+    def create_data(self, args):
```
annalena-k, line 333: Why do we need two times the same method, but with different functionalities? (see line 330)

srashtig:
Hi guys, is there a plan to merge this with the main branch anytime soon? If yes, it'd be helpful for the lensing studies, too. Perhaps I can also try to resolve the conflicts if it'll help.

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
...
-                f_max=self.maximum_frequency,
-                delta_f=1 / self.duration,
-            )
-            strain = domain.update_data(strain)
-            asd = domain.update_data(asd, low_value=1.0)
-
-            # Dingo expects data to have trigger time 0, so we apply a cyclic time shift
-            # by the post-trigger duration.
-            strain = domain.time_translate_data(strain, self.post_trigger_duration)
+            if np.max(np.isnan(asd)):
```
annalena-k, line 342: Easier to understand:
```suggestion
            if np.any(np.isnan(asd)):
```

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
...
+                        self.sampling_frequency,
+                    )
+                    psd = PowerSpectralDensity(
+                        frequency_array=frequency_array, psd_array=psd_array
+                    )
+                    ifo.power_spectral_density = psd
+                else:
+                    logger.critical(
+                        f"""Nan encountered in strain data for PSD estimation for detector {ifo.name}. 
+                    Specify --shift-segment-for-psd-generation-if-nan to shift PSD segement to an earlier time without Nans. """
```
annalena-k, line 364: Can you please run `black` on all files before merging to correct the formatting of long lines like this one?

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
...
+                    )
+                    raise
```
annalena-k, line 366: Is there a reason, you didn't raise an exception, e.g. `ValueError`?
This looks really strange in the output file (see screenshot)
image

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
...
-            strain = domain.time_translate_data(strain, self.post_trigger_duration)
+            if np.max(np.isnan(asd)):
+                if args.shift_segment_for_psd_generation_if_nan:
+                    window = {
+                        "type": "tukey",
+                        "roll_off": self.tukey_roll_off,
+                        "T": self.duration,
+                        "f_s": self.sampling_frequency,
+                    }
+                    psd_array = download_psd(
```
annalena-k, line 350: I noticed that this function uses `print` internally, instead of writing to the logger.
Not sure whether this can be addressed immediately?

dingo/pipe/data_generation.py
```diff
@@ -162,47 +194,187 @@ def __init__(self, args, unknown_args, create_data=True):
...
+                        ifo.name,
+                        self.start_time + self.psd_start_time,
```
annalena-k, line 352: Summary of a discussion with Nihar: It is fine to add `self.psd_start_time` here because it is defined as a negative value, e.g. `self.psd_start_time = 1024`. This means that `self.start_time + self.psd_start_time` is indeed smaller than `self.start_time` (as it should be). Might be good to include a comment here?

dingo/core/nn/nsf.py
nihargupte-ph: Done

dingo/gw/injection.py
nihargupte-ph: I think the seed is just being used for the noise realization (next few lines) so numpy only should be ok

dingo/pipe/data_generation.py
nihargupte-ph: Done

dingo/pipe/data_generation.py
nihargupte-ph: In the end multibanding was merged first, but I've updated this to be compatible.

dingo/pipe/data_generation.py
```diff
@@ -150,7 +182,7 @@ def __init__(self, args, unknown_args, create_data=True):
         # self.plot_injection = args.plot_injection
 
         if create_data:
-            # Added for dingo so that create_data runs. TODO: enable injections
+            # TODO what happens if we remove this??
```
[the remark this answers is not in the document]
nihargupte-ph: Removed this comment, I"m not sure what it was for...

dingo/pipe/data_generation.py
nihargupte-ph: Not sure, I've deleted the redundant one

dingo/pipe/data_generation.py
nihargupte-ph: done

dingo/pipe/data_generation.py
nihargupte-ph: True, I've added a ValueError now

dingo/pipe/data_generation.py
nihargupte-ph: True, I've added a comment

dingo/pipe/data_generation.py
nihargupte-ph: Fixed

dingo/pipe/data_generation.py
nihargupte-ph: Done

dingo/pipe/sampling.py
```diff
@@ -180,7 +189,9 @@ def density_recovery_settings(self, settings):
     #     return os.path.relpath(result_dir)
 
     def run_sampler(self):
-        if self.gnpe and self.recover_log_prob:
+        self.dingo_sampler.event_metadata = self.event_metadata
+        if self.gnpe and self.recover_log_prob and not self.zero_noise:
+            self.dingo_sampler.context = self.context.data
```
annalena-k, line 194: This code only works for two options:
1. If you train a GNPE network, i.e. `self.gnpe = True` and `self.recover_log_prob = True`.
2. If you want to run injections

What about running dingo pipe with an NPE network? Since there is no `else` option, `self.dingo_sampler.context` is never set, so `run_sampler` would not work.

dingo/pipe/sampling.py
```diff
@@ -134,9 +146,6 @@ def _load_sampler(self):
             self.gnpe = False
             self.dingo_sampler = GWSampler(model=model)
 
-        self.dingo_sampler.context = self.context
```
annalena-k, line 137: I just want to point out that `GWSampler.context` is a setter which means that some parts of the class might be changed based on the provided data. I think this is not a problem for you, but the code update I pushed to the masking PR includes a line [here](https://github.com/dingo-gw/dingo/blob/18ee73afa87c26036bdc1e03b8e9f867d4ae683b/dingo/pipe/sampling.py#L133) which re-initializes the transforms based on the frequency ranges provided.

If you move the lines here into the `run_sampler` method, we need to copy the lines from the masking PR into all the options you allow in `run_sampler` which isn't really elegant.
Why did you decide not to use an `if` here to only set the context and event metadata if it is not an injection?

dingo/pipe/sampling.py
```diff
@@ -193,9 +204,51 @@ def run_sampler(self):
                 **self.density_recovery_settings,
             )
 
-        self.dingo_sampler.run_sampler(self.num_samples, batch_size=self.batch_size)
-        self.dingo_sampler.to_hdf5(label=self.label, outdir=self.result_directory)
+        # Training unconditional density estimator if zero noise
+        elif self.zero_noise:
+            n_training_samples = 1_000_000
```
annalena-k, line 209: Do we really want to hard-code this value?

dingo/pipe/sampling.py
```diff
@@ -193,9 +204,51 @@ def run_sampler(self):
...
+                    "asds": self.context.data["asds"]
+                }
+                self.dingo_sampler.context = context
+                self.dingo_sampler.run_sampler(
+                    int(n_training_samples / self.num_noise_realizations),
+                    batch_size=self.batch_size,
+                )
+                samples_list.append(self.dingo_sampler.samples)
 
+                logger.info(
```
annalena-k, line 223: Why is this logging statement within the loop over multiple noise realizations? It will be printed for every considered noise realization?

dingo/pipe/sampling.py
annalena-k: To me, it seems like a better idea to have the injections in the `if` condition and all other options in the `else` (GNPE, NPE, FMPE,....)

dingo/pipe/sampling.py
```diff
@@ -37,11 +39,22 @@ def __init__(self, args, unknown_args):
...
+        self.model_init = args.model_init
+        self.zero_noise = args.zero_noise
+        self.injection = args.injection_dict is not None
+        self.recover_log_prob = args.recover_log_prob
+        if self.zero_noise:
+            self.num_noise_realizations = args.num_noise_realizations
+            self.recover_log_prob = False
+        else:
+            self.num_noise_realizations = 1
+            self.recover_log_prob = args.recover_log_prob
```
annalena-k, line 57: `self.recover_log_prob` actually gets overwritten again in line 67.

dingo/pipe/sampling.py
```diff
@@ -37,11 +39,22 @@ def __init__(self, args, unknown_args):
...
         self.result_format = args.result_format
 
-        # Event file to run on
+        # Event files to run on
         self.event_data_file = args.event_data_file
 
         # Choices for running
         self.detectors = args.detectors
+        self.model = args.model
+        self.model_init = args.model_init
```
annalena-k, line 48: This is duplicate code from lines 61 - 66

annalena-k (changes requested):
There appear to be some errors from merging main into this branch.
Have you tried whether the injections code automatically works with MFD injections?

dingo/pipe/sampling.py
annalena-k: We have to make sure that `self.dingo_sampler.context = self.context.data` is called in any case because otherwise there's no data stored as a condition and the sampling would fail.

dingo/pipe/sampling.py
```diff
@@ -193,9 +204,51 @@ def run_sampler(self):
...
+
+                nde_sampler = GWSampler(model=unconditional_flow)
+                nde_sampler.metadata = self.dingo_sampler.metadata
+                nde_sampler.run_sampler(
+                    num_samples=self.num_samples, batch_size=self.batch_size
+                )
+                self.dingo_sampler = nde_sampler
+
+        # run the sampler
+        self.dingo_sampler.run_sampler(
```
annalena-k, line 247: Why do we run the sampler again for the last noise realization in the loop?

dingo/pipe/sampling.py
```diff
@@ -193,9 +204,51 @@ def run_sampler(self):
...
+            num_samples=self.num_samples, batch_size=self.batch_size
+        )
+
+        self.dingo_sampler.to_hdf5(label=self.label, outdir=self.result_directory)
```
annalena-k, line 251: We only save the `dingo_sampler` for the last noise realization in an hdf5 because `self.dingo_sampler` gets overwritten in line 244.
I guess we only need to store one sampler for all noise realizations. But why do we then overwrite `self.dingo_sampler = nde_sampler` in the noise realization loop? Why not outside the loop?

annalena-k: But why do we not store the sampler outputs for different noise realizations? Why is there not `.to_hdf5()` within the loop over different noise realizations?

nihargupte-ph: See updated code, the previous one was wrong, the training of the unconditional density estimator shouldn't not be happening in the loop of noise realizations but rather outside with the set of pooled samples.

dingo/pipe/sampling.py
nihargupte-ph: See below, answer: we shouldn't!

dingo/pipe/sampling.py
nihargupte-ph: Good catch, didn't spot that when I was merging

dingo/pipe/sampling.py
nihargupte-ph: It shouldn't be (see update)

dingo/pipe/sampling.py
nihargupte-ph: Updated, I included an argument in `parser.py`

dingo/pipe/sampling.py
nihargupte-ph: Updated, Alex mentioned this to me as well, I just forgot to include it.

dingo/pipe/sampling.py
nihargupte-ph: I edited it now to move this back to to _load_sampler method and only reset the context when doing zero-noise injections.

dingo/pipe/sampling.py
annalena-k: Thanks, that makes more sense!

annalena-k (approved):
Looks good to me, it makes sense now.
I haven't tested the code myself though.

stephengreen (changes requested):
Thanks for preparing this. What's the rationale for not leveraging the existing `bilby_pipe` injections code? I think it should be possible to enable injections just by uncommenting some of the code taken over from `bilby_pipe`, essentially allowing the various cases in `create_data` to execute.

Leveraging `bilby_pipe` would make for much more maintainable code. It would easily allow for injections into Gaussian noise, real noise (not present in the PR), or zero noise (not what you want, but maybe useful for experimenting), but one may want to augment it with an averaging option for zero-noise-like injections. It would also allow to inject NR waveforms.

Of course, using `bilby_pipe` in this way would mean using Bilby for all of the waveforms, which would mean not using the Dingo injections code (`dingo.gw.injection`). This would limit injections to those that Bilby can already produce (e.g., not allow for modifications to waveform models, or data generation processes). For this, one would have to use the Dingo injections API directly (i.e., not `dingo_pipe`) or include an option for Dingo injections.

PR #211, opened 2023-07-11
Compute SNR posteriors

mpuerrer:
* Implement SNR posteriors for standard likelihood
* `log_likelihood()` now returns 3 floats (logL, rho_opt, rho_coh) and similarly `log_likelihood_multi()` will now return arrays of these quantities.

TODO:
  * implement the SNRs for the marginalized likelihoods
  * add tests

### DIFF
--- a/dingo/core/result.py
+++ b/dingo/core/result.py
@@ -278,7 +278,7 @@ class Result(DingoDataset):
 
         print(f"Calculating {len(theta)} likelihoods.")
         t0 = time.time()
-        log_likelihood = self.likelihood.log_likelihood_multi(
+        log_likelihood, rho_opt, rho_coh = self.likelihood.log_likelihood_multi(
             theta, num_processes=num_processes
         )
         print(f"Done. This took {time.time() - t0:.2f} seconds.")
@@ -286,6 +286,8 @@ class Result(DingoDataset):
         self.log_noise_evidence = self.likelihood.log_Zn
         self.samples["log_prior"] = log_prior
         self.samples.loc[valid_samples, "log_likelihood"] = log_likelihood
+        self.samples.loc[valid_samples, "rho_opt"] = rho_opt
+        self.samples.loc[valid_samples, "rho_coh"] = rho_coh
         self._calculate_evidence()
 
     def _calculate_evidence(self):
--- a/dingo/gw/likelihood.py
+++ b/dingo/gw/likelihood.py
@@ -247,6 +247,8 @@ class StationaryGaussianGWLikelihood(GWSignal, Likelihood):
         Returns
         -------
         log_likelihood: float
+        rho_opt: float
+        rho_coh: float
         """
 
         # Step 1: Compute whitened GW strain mu(theta) for parameters theta.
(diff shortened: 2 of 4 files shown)


stephengreen:
I have updated this PR with several changes:
* Made the code in `core.likelihood` more generic by not referring explicitly to GW quantities. The generic `log_likelihood` method now has an argument flag `include_supplemental`. If `False` (default), the method acts as before, returning a scalar log likelihood. If `True`, it returns a tuple containing the log likelihood, and the supplemental information as a dict.
* For GWs, the supplemental information includes the per-detector optimal and matched-filter SNRs.
* I modified the `inner_product` function to assume that the frequency axis is last rather than first. This allowed to streamline the code for the calibration-marginalized log likelihood, so I could unify it with the ordinary log likelihood.
* There are now unit tests comparing the log likelihoods from Bilby and Dingo, for IMRPhenomXPHM, SEOBNRv4PHM, and SEOBNRv5PHM models.
  - These create an injection at fixed parameters, and then calculate the log likelihoods at random parameter values. The threshold for comparison is 0.1, and this usually passes, but occasionally fails with slightly higher deviations. I'm not sure the cause of these small deviations.

dingo/core/result.py
```diff
@@ -278,14 +278,16 @@ def importance_sample(self, num_processes: int = 1, **likelihood_kwargs):
 
         print(f"Calculating {len(theta)} likelihoods.")
         t0 = time.time()
-        log_likelihood = self.likelihood.log_likelihood_multi(
-            theta, num_processes=num_processes
+        log_likelihood, supplemental = self.likelihood.log_likelihood_multi(
```
mpuerrer, line 281: For clarity `supplemental` -> `supplemental_dict` here and below in line 289.

dingo/gw/likelihood.py
```diff
@@ -192,13 +192,29 @@ def log_likelihood(self, theta):
...
-            return self._log_likelihood_calibration_marginalized(theta)
+            result = self._log_likelihood(theta)
         else:
-            return self._log_likelihood(theta)
+            result = self._log_likelihood(theta)
+
+        # We assume that the _log_likelihood* methods return either a 2-tuple (log_likelihood, supplement), or just a
+        # log_likelihood scalar. If just a scalar, but supplemental information was requested, return an empty dict.
+
+        if include_supplemental:
```
mpuerrer, line 206: Why would some of the log_likelihood functions not return the supplement? If they did not compute the SNRs (or other info) for some reason. But then, they could return an empty dict and this code could be simplified.

dingo/gw/likelihood.py
```diff
@@ -247,22 +265,53 @@ def _log_likelihood(self, theta):
...
+            ifo: inner_product(d[ifo], mu_ifo) for ifo, mu_ifo in mu.items()
+        }
+
+        # Calculate likelihood.
+
+        # First sum over detectors for each draw.
+        rho2opt = np.sum(list(rho2opt_detector.values()), axis=0)
+        kappa2 = np.sum(list(kappa2_detector.values()), axis=0)
+
+        if isinstance(rho2opt, np.ndarray):
```
mpuerrer, line 295: Add another comment that this is the calibration marginalization case.

dingo/gw/likelihood.py
```diff
@@ -247,22 +265,53 @@ def _log_likelihood(self, theta):
...
+        rho_opt_detector = {
+            f"{ifo}_optimal_snr": np.mean(v ** (1 / 2))
+            for ifo, v in rho2opt_detector.items()
+        }
+        rho_mf_detector = {
+            f"{ifo}_matched_filter_snr": np.mean(v / rho2opt_detector[ifo] ** (1 / 2))
+            for ifo, v in kappa2_detector.items()
+        }
+
+        return log_likelihood, rho_opt_detector | rho_mf_detector
```
mpuerrer, line 314: For added readability consider adding a line `rho_detectors_dict = rho_opt_detector | rho_mf_detector` and returning `rho_detectors_dict`.

tests/gw/test_gw_likelihood.py
```diff
@@ -0,0 +1,233 @@
...
+        "luminosity_distance": 1000.0,
+        "geocent_time": 0.0,
+        "ra": 1.0,
+        "dec": 2.0,
+        "psi": 2.5,
+    }
+
+
+@pytest.fixture  # (params=["IMRPhenomXPHM", "SEOBNRv4PHM", "SEOBNRv5PHM"])
+def injection(approximant):
```
mpuerrer, line 47: Consider renaming the function to something like `prepare_injection()` to make the later usage `injection.injection` clearer.

mpuerrer (commented):
Looks good. I made some minor comments.

PR #212, opened 2023-08-04
Fix bug in conversion of spin angles to Cartesian spins for new waveform interface (SEOBNRv5PHM)

nihargupte-ph:
For v5 waveforms the spins were not being converted properly. This was due to the fact that the masses did not have the correct dimensions in the `bilby_to_lalsimulation_spins` function.

### DIFF
--- a/dingo/gw/waveform_generator/waveform_generator.py
+++ b/dingo/gw/waveform_generator/waveform_generator.py
@@ -16,7 +16,7 @@ try:
     from lalsimulation.gwsignal.models import (
         gwsignal_get_waveform_generator as new_interface_get_waveform_generator,
     )
-except ImportError:
+except:
     pass
 
 from bilby.gw.conversion import (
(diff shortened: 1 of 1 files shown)


dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -16,7 +16,7 @@
     from lalsimulation.gwsignal.models import (
         gwsignal_get_waveform_generator as new_interface_get_waveform_generator,
     )
-except ImportError:
+except:
```
stephengreen, line 19: This should raise an ImportError, no? A generic exception is not good practice.
```suggestion
except ImportError:
```

stephengreen (changes requested):
Thanks Nihar, it was great to find this! One small change requested.

stephengreen (approved):

PR #213, opened 2023-08-18
Parallelize post-processing change of spin-conversion-phase

stephengreen:
Dingo and Bilby use different conventions for using the phase parameter in converting between PE and Cartesian spins. (Dingo typically fixes it to 0, whereas Bilby uses the actual phase parameter.) This makes comparing final posteriors difficult, hence we have a method `gw.result.Result.get_samples_bilby_phase` to convert. In turn, this uses `gw.conversion.spin_conversion.change_spin_conversion_phase`.

This update simply parallelizes this function, since it is somewhat slow.

**Note:** This does not speed up the property `gw.result.Result.pesummary_samples`, since it is not clear how to pass the `num_processes` argument to a property. It would be good to accelerate also this property, since this is ultimately called by PESummary.


nihargupte-ph:
I needed this PR to post-process the results from some of the recent LVK runs. I adjusted the result class just now to change the property `pesummary_samples` to `get_pesummary_samples` where one can then specify `num_processes`. The pesummary package needs to be updated to handle this case.

nihargupte-ph (approved):
I tested it locally and it seems to work.

Issue #216, opened 2023-09-19
Unable to use Dingo Pipe with 1 detector networks due to bad size of nflows training array


nihargupte-ph:
When doing inference with a 1 detector network, one needs to train an unconditional density estimator on the set of 1D GNPE samples. However, for some reason nflows does not like this as it thinks the training vector is the wrong shape. This occurs during the sampling step of dingo_pipe. The error message is the following: 

```
[...]

PR #218, opened 2023-09-20
Enable Support for one detector networks

nihargupte-ph:
This pull request enables support for 1 detector networks. 

There are two current issues with one detector flows for Dingo. 

1). https://github.com/dingo-gw/dingo/issues/216. Summarized, there is an issue when using a coupling layer with a 1D distribution with no context as there is no input to the neural network. This causes nflows to error out. 
> This PR fixes this by using an autoregressive flow to learn the 1D density of GNPE samples. 

2). Currently, there is no way to pass no context during gibbs sampling 
> There is a simple fix to this which is in lines 503 of dingo.core.samplers.py. 

This PR fixes the issues. It has been tested on GW190620 which is a L1 only event. It achieves a sample efficiency of 6.57% with 32_000 effective samples in this case.


dingo/gw/inference/inference_pipeline.py
stephengreen, line 1: Are you using the `get_event_data()` function in this file? It's only called by `analyze_event()`, not `dingo_pipe`.

dingo/gw/inference/inference_pipeline.py
```diff
@@ -241,6 +244,17 @@ def prepare_log_prob(
...
+    # If there is only one parameter, and no context, we cannot use a
+    # coupling transform. In this case, we use an autoregressive transform
+    # for the density estimator
+    if len(sampler.gnpe_proxy_parameters) == 1:
+        print(
+            "Using autoregressive transform for density estimator since there is only 1 GNPE proxy."
+        )
+        nde_settings["model"]["base_transform_kwargs"][
+            "base_transform_type"
+        ] = "rq-autoregressive"
```
stephengreen, line 257: I think choosing these options might be better done in `dingo_pipe`, where the `density_recovery_settings` get chosen. The `prepare_log_prob()` function should just take those settings and run them.

dingo/core/samplers.py
```diff
@@ -492,10 +492,19 @@ def _run_sampler(
...
+
             with torch.no_grad():
-                y, log_prob = self.model.model.sample_and_log_prob(
-                    x["data"], x["context_parameters"]
-                )
+                # If we are running with one detector, we need not pass extra context to
+                # the network. This is because there is a global shift which is applied to the
+                # geocent_time according to the time proxies.
+                # Here len(x["extrinsic_parameters"]) == 2, is only the 1 detector proxy and
+                # the geocent time.
```
stephengreen, line 501: `core.samplers` should not have to know anything about gravitational waves. Maybe there is another way to do this...

stephengreen: So the worry is that `x["context_parameters"]` is empty?

PR #220, opened 2023-09-25
Add plot_density option to plot_corner_multi

stephengreen:
This is useful for when plotting posteriors with different numbers of samples. Internally it redefines the weight arrays to have unit sum.


nihargupte-ph (approved):
I tried it on a couple posteriors, seems to work fine.

stephengreen:
Based on your comment @nihargupte-ph it sounds like this PR does not fix the plotting issue?

PR #223, opened 2023-10-19
Set `spin_conversion_phase = 0.0` for NPE and GNPE examples

stephengreen:
These examples fail when sampling the synthetic phase since they defaulted to `spin_conversion_phase = None`.


stephengreen:
We should also document this setting (in the waveform dataset generation part of the docs) and possibly print a warning when training a phase-marginalized network with `spin_conversion_phase != 0.0`.

stephengreen:
Updated the docs.

PR #229, opened 2023-11-03
Update pyproject.toml dependencies

stephengreen:
Adding the following dependencies, which are explicitly used in the code:

* requests
* asimov
* matplotlib

Removing:

* cryptography
* pyopenssl

@nihargupte-ph I think you added `cryptography` and `pyopenssl` in #121, where you put a fixed version. Is this still needed?


stephengreen:
`asimov` also depends on `htcondor`, which on pypi only exists for Linux. Maybe we should not have `asimov` as a dependency then. In that case, we should handle the imports appropriately.

@max-dax @jonaswildberger What do you think?

mpuerrer:
I think we should avoid requiring dependencies such as `htcondor` because it is not needed for all use cases of dingo and will make it hard for people to try out the code. Such dependencies should be optional and could be listed in the README and docs.

stephengreen:
Good point, I removed Asimov from the dependencies.

max-dax (approved):

PR #231, opened 2023-11-09
added support for 1 detector networks

nihargupte-ph:
This pull request enables support for 1 detector networks.

There are two current issues with one detector flows for Dingo.

1). https://github.com/dingo-gw/dingo/issues/216. Summarized, there is an issue when using a coupling layer with a 1D distribution with no context as there is no input to the neural network. This causes nflows to error out.

This PR fixes this by using an autoregressive flow to learn the 1D density of GNPE samples.

2). Currently, there is no way to pass no context during gibbs sampling

There is a simple fix to this which is in lines 503 of dingo.core.samplers.py.

This PR fixes the issues. It has been tested on GW190620 which is a L1 only event. It achieves a sample efficiency of 6.57% with 32_000 effective samples in this case.

### DIFF
--- a/dingo/core/nn/nsf.py
+++ b/dingo/core/nn/nsf.py
@@ -2,6 +2,7 @@
 Implementation of the neural spline flow (NSF). Most of this code is adapted
 from the uci.py example from https://github.com/bayesiains/nsf.
 """
+
 import copy
 
 import torch
@@ -218,7 +219,7 @@ class FlowWrapper(nn.Module):
         if self.embedding_net is not None:
             x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
         if len(x) > 0:
-            sample = self.flow.sample(num_samples)
+            sample = self.flow.sample(num_samples, x)
             sl = list(range(sample.ndim - 1))
             return torch.squeeze(sample, dim=sl)
         else:
--- a/dingo/core/nn/nsf.py
+++ b/dingo/core/nn/nsf.py
@@ -218,21 +218,27 @@ class FlowWrapper(nn.Module):
         if self.embedding_net is not None:
             x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
         if len(x) > 0:
-            return torch.squeeze(self.flow.sample(num_samples, x))
+            sample = self.flow.sample(num_samples)
+            sl = list(range(sample.ndim - 1))
+            return torch.squeeze(sample, dim=sl)
         else:
             # if there is no context, omit the context argument
-            return torch.squeeze(self.flow.sample(num_samples))
+            sample = self.flow.sample(num_samples)
+            sl = list(range(sample.ndim - 1))
+            return torch.squeeze(sample, dim=sl)
 
     def sample_and_log_prob(self, *x, num_samples=1):
         if self.embedding_net is not None:
             x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
         if len(x) > 0:
             sample, log_prob = self.flow.sample_and_log_prob(num_samples, x)
-            return torch.squeeze(sample), torch.squeeze(log_prob)
+            sl = list(range(sample.ndim - 1))
+            return torch.squeeze(sample, dim=sl), torch.squeeze(log_prob, dim=sl)
         else:
             # if there is no context, omit the context argument
             sample, log_prob = self.flow.sample_and_log_prob(num_samples)
-            return torch.squeeze(sample), torch.squeeze(log_prob)
+            sl = list(range(sample.ndim - 1))
+            return torch.squeeze(sample, dim=sl), torch.squeeze(log_prob, dim=sl)
 
     def forward(self, y, *x):
         if len(x) > 0:
--- a/dingo/core/nn/nsf.py
+++ b/dingo/core/nn/nsf.py
@@ -219,27 +219,19 @@ class FlowWrapper(nn.Module):
         if self.embedding_net is not None:
             x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
         if len(x) > 0:
-            sample = self.flow.sample(num_samples, x)
-            sl = list(range(sample.ndim - 1))
-            return torch.squeeze(sample, dim=sl)
+            return self.flow.sample(num_samples, x)
         else:
             # if there is no context, omit the context argument
-            sample = self.flow.sample(num_samples)
-            sl = list(range(sample.ndim - 1))
-            return torch.squeeze(sample, dim=sl)
+            return self.flow.sample(num_samples)
 
     def sample_and_log_prob(self, *x, num_samples=1):
         if self.embedding_net is not None:
             x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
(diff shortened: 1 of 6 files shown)


dingo/core/nn/nsf.py
```diff
@@ -218,21 +218,27 @@ def sample(self, *x, num_samples=1):
         if self.embedding_net is not None:
             x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
         if len(x) > 0:
-            return torch.squeeze(self.flow.sample(num_samples, x))
+            sample = self.flow.sample(num_samples)
+            sl = list(range(sample.ndim - 1))
+            return torch.squeeze(sample, dim=sl)
```
stephengreen, line 223: I'm trying to understand the issue here. Is the original reason for the squeeze to drop the leading dimension of size 1 when `num_samples=1`?  And the problem occurs for the single-detector time network because the output is a single time, leads to dropping the last dimension? If so, a simpler fix could be
```suggestion
            sample = self.flow.sample(num_samples)
            return torch.squeeze(sample, dim=0)
```

Line 221 also looks like an error, no? Don't we need to pass `x` to `flow.sample`?

dingo/core/samplers.py
```diff
@@ -492,10 +492,14 @@ def _run_sampler(
...
             with torch.no_grad():
-                y, log_prob = self.model.model.sample_and_log_prob(
-                    x["data"], x["context_parameters"]
-                )
+                if "context_parameters" not in x.keys():
+                    y, log_prob = self.model.model.sample_and_log_prob(x["data"])
+                else:
+                    y, log_prob = self.model.model.sample_and_log_prob(
+                        x["data"], x["context_parameters"]
+                    )
```
stephengreen, line 502: I would swap the order of the `if`/`else` so you have fewer negatives, and also you don't need the `.keys()`:
```suggestion
                if "context_parameters" in x:
                    y, log_prob = self.model.model.sample_and_log_prob(
                        x["data"], x["context_parameters"]
                    )
                else:
                    y, log_prob = self.model.model.sample_and_log_prob(x["data"])
```

stephengreen (changes requested):
I had a couple questions on the modification to the `FlowWrapper`, but otherwise things look good.

dingo/core/nn/nsf.py
nihargupte-ph: I think you mostly described the issue. The only additional thing is that by default, sample has an extra dimension. So e.g. if you have `num_samples = 50` and a output dimension of 2, the size of `sample` is actually `(50, 1,  2)`. I think this is important during training where the middle dimension may be something other than 1 to account for the batch. So then in general it is better to squeeze all dims except the last one. 

Yes line 221 should also be passing `x` to `flow.sample`. This has been updated in commit `45a2b17`

stephengreen: During training, don't we ever only evaluate the `log_prob`?

stephengreen: Okay I stepped through the code, and indeed I see the shape you describe. I would have thought the first dimension corresponded to the batch though, since `x` is of shape `(50, 128)`, with `50` the (repeated) batch dimension. So I don't see why `nflows` is inserting the singleton dimension. Anyway, `torch.squeeze` eliminated it.

With a 1D parameter space (e.g., the initialization network for 1 detector) the squeezing would also drop the last dimension, which would be problematic. So your fix resolves this.

Turning to the first dimension (corresponding to the number of samples), the present version (and previous) gives an error when requesting just one sample. This error is of the same nature as above, since the leading dimension is eliminated by the squeeze, which the sampler class does not expect. A fix for this would be to return `torch.squeeze(sample, dim=1)`, if we are convinced that works in all cases.

Thoughts @max-dax @nihargupte-ph ?

stephengreen:
In testing, I noticed a bug in the `Injection` class, where the random injection had different datatypes (`float` or `np.float`) for different parameters. I'm not sure why this occurred, but likely it comes from Bilby. I am now casting to `float`.

If some parameters were complex, this would cause a problem, but I don't believe that is ever the case.

dingo/core/nn/nsf.py
stephengreen: Looking more closely, the singleton dimension refers to the number of samples requested from `Flow.sample()`. There are two ways in which this is called:
1. For conditional NPE flows and GNPE models, we call this with `num_samples=1` (setting the requested number of samples instead via the leading dimension of the context). In this case, we want to squeeze `dim=1`.
2. For an unconditional flow, we have no context, so we set the number of samples via  `num_samples>1`. In this case, I believe `x` will be of shape `(num_samples, param_dim)`, and we don't want to squeeze anything. In fact, if we did my suggestion of squeezing `dim=1`, we would run into a problem with the unconditional 1D flow for single-detector networks.

I think what we should do then is
```
if sample.ndim == 3:
    sample = torch.squeeze(sample, dim=1)
return sample
```
What do people think?

stephengreen: Perhaps we could alternatively make use of the existing `if/else` statement that refers to the presence of the context.

stephengreen: Okay, how about this @max-dax @nihargupte-ph 

```
    def sample(self, *x, num_samples=1):
        if self.embedding_net is not None:
            x = torchutils.forward_pass_with_unpacked_tuple(self.embedding_net, x)
        if len(x) > 0:
            sample = self.flow.sample(num_samples, x)
            assert sample.ndim == 3
            if num_samples == 1:
                return torch.squeeze(sample, dim=1)
            else:
                return torch.squeeze(sample, dim=0)
        else:
            # if there is no context, omit the context argument
            return self.flow.sample(num_samples)
```
and similarly for `sample_and_log_prob()`.

nihargupte-ph: I haven't tested it so maybe it works. But if you set `dim=(1 if num_samples==1 else 0)` then wouldn't this not drop the middle dimension?

But in any case, if we know the meaning of the last dimension, and we know it should never get squeezed, isn't the most minimal change to just not squeeze it? It keeps the rest of the logic the same.

stephengreen: I've taken a new approach to this, completely removing the squeezing from the `FlowWrapper`, since it is hard to ascertain the various use cases from within that class. This makes the `FlowWrapper` methods much more straightforward.

Instead, the squeezing is done in the `_run_sampler()` methods of the `Sampler` and `GNPESampler` classes, after calling `FlowWrapper.sample_and_log_prob()`. These classes know what they are trying to do, so they know which dimension to squeeze.

One additional optimization I made is to pass the data through the embedding network only once per batch for NPE networks, rather than replicating it `num_samples` times. We instead use the `num_samples` argument of `FlowWrapper.sample_and_log_prob()` to set the number of samples. (This can't be done for GNPE networks because each data sample is transformed differently according to the parameters.)

I had to update one unit test so that it continues to pass. I also tested sampling form NPE and GNPE networks, and it works. It would be good to test single detector networks, as well as unconditional flows (e.g., by running dingo_pipe on an event).

nihargupte-ph: I checked sampling with a 1D network using dingo_pipe. I am able to sample and train the unconditional flow.

stephengreen (approved):

PR #233, opened 2023-11-20
Fix bug related to random seeds

max-dax:
This fixes a bug, where each worker was initialised with the same random seed due to a change made in [bilby](https://git.ligo.org/lscsoft/bilby/-/merge_requests/1273). 

This bug likely lead to identical samples for extrinsic parameters, which corrupted the neural network training in the following way. The network could just memorise the parameters from the previous workers, without looking at the strain data. With `N` data loader workers, this resulted in a pattern where the loss would rapidly decrease over `N` iterations, as the network memorised the parameters. In iteration `N+1`, the loss steeply increased (as all workers would generate new random numbers), and then rapidly decrease again until iteration `2N`.

### DIFF
--- a/dingo/core/utils/torchutils.py
+++ b/dingo/core/utils/torchutils.py
@@ -8,9 +8,12 @@ import bilby
 
 
 def fix_random_seeds(_):
-    """Util function to set random seeds when using multiple workers for Dataloader."""
+    """Utility function to set random seeds when using multiple workers for DataLoader."""
     np.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
-    bilby.core.utils.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
+    try:
+        bilby.core.utils.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
+    except AttributeError:  # In case using an old version of Bilby.
+        pass
 
 
 def get_activation_function_from_string(activation_name: str):
--- a/dingo/core/utils/torchutils.py
+++ b/dingo/core/utils/torchutils.py
@@ -4,6 +4,13 @@ import torch.nn as nn
 from torch.nn import functional as F
 from torch.utils.data import DataLoader
 from typing import Union, Tuple, Iterable
+import bilby
+
+
+def fix_random_seeds(_):
+    """Util function to set random seeds when using multiple workers for Dataloader."""
+    np.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
+    bilby.core.utils.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
 
 
 def get_activation_function_from_string(activation_name: str):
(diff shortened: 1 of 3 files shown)


dingo/gw/training/train_builders.py
```diff
@@ -271,9 +271,7 @@ def build_svd_for_embedding_network(
         wfd,
         batch_size=batch_size,
         num_workers=num_workers,
-        worker_init_fn=lambda _: np.random.seed(
-            int(torch.initial_seed()) % (2 ** 32 - 1)
-        ),
+        worker_init_fn=fix_random_seeds,
```
stephengreen, line 274: I think you need to import this, no?

max-dax: It's imported via `from dingo.core.utils import *`. This is maybe not good style, but not part of this PR.

stephengreen: Ah, I missed that1

stephengreen:
It looks good to me, but needs to be tested @nihargupte-ph

nihargupte-ph:
Looks fine to me. I ran a model starting at epoch 200 through the debugger. I am not sure why but the loss is starting higher than I expected (12). Nonetheless, I find that the loss decreases after 32 iterations (the number of workers) but does not again increase. I attach a log file here. 
[log.txt](https://github.com/dingo-gw/dingo/files/13417558/log.txt)

stephengreen:
> Looks fine to me. I ran a model starting at epoch 200 through the debugger. I am not sure why but the loss is starting higher than I expected (12). Nonetheless, I find that the loss decreases after 32 iterations (the number of workers) but does not again increase. I attach a log file here. [log.txt](https://github.com/dingo-gw/dingo/files/13417558/log.txt)

The reason for this could be that the random seed for each worker is reset to the same value at the beginning of each epoch. This might be the case if the DataLoader creates new workers each epoch, although I don't know why the seed would be the same, unless Bilby is setting it somewhere. In any case, this would in principle allow the network to overfit to the training data from epoch to epoch, and could result in the behavior you are seeing.

stephengreen:
Docs are failing to build? Could someone investigate this? We will need docs to build before we can merge.

nihargupte-ph:
> Docs are failing to build. Could someone @nihargupte-ph @max-dax investigate this? We will need docs to build before we can merge.

So I think it's because the right version of docutils isn't not being used in the CI script. This line,

```
pip install --force-reinstall "docutils<0.17"
```

should be replaced with 

```
python -m pip install --force-reinstall "docutils==0.18.1"
```

dingo/core/utils/torchutils.py
```diff
@@ -4,6 +4,13 @@
 from torch.nn import functional as F
 from torch.utils.data import DataLoader
 from typing import Union, Tuple, Iterable
+import bilby
+
+
+def fix_random_seeds(_):
+    """Util function to set random seeds when using multiple workers for Dataloader."""
+    np.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
+    bilby.core.utils.random.seed(int(torch.initial_seed()) % (2 ** 32 - 1))
```
stephengreen, line 13: This should probably be placed in a `try`/`except` statement in case an older version of Bilby is used.

stephengreen (approved):

stephengreen:
> should be replaced with
> 
> ```
> python -m pip install --force-reinstall "docutils==0.18.1"
> ```

I just removed this line and it compiles fine. Thanks.

PR #235, opened 2024-02-01
Fixed bug in NotImplementedError message

annalena-k:
This PR fixes a bug in a NotImplementedError message.

When running importance sampling without phase marginalization and a waveform model other than IMRPhenomXPHM, a `NotImplementedError` should be hit in `generate_FD_modes_LO()` of `dingo/gw/waveform_generator/waveform_generator.py` .

However, the error message itself throws the error
`TypeError: in method 'GetApproximantFromString', argument 1 of type 'char const *'*`
in line https://github.com/dingo-gw/dingo/blob/fa032dbfb3494118868fa9ac5efd88992fbd1416/dingo/gw/waveform_generator/waveform_generator.py#L694

This error occurs because `self.approximant` is an integer while the function `LS.GetApproximantFromString(self.approximant)` only takes strings/char-type inputs.

The command is replaced by `self.approximant_str` and the error message now prints correctly.

### DIFF
--- a/dingo/gw/waveform_generator/waveform_generator.py
+++ b/dingo/gw/waveform_generator/waveform_generator.py
@@ -691,7 +691,7 @@ class WaveformGenerator:
             return hlm_fd, iota
         else:
             raise NotImplementedError(
-                f"Approximant {LS.GetApproximantFromString(self.approximant)} not "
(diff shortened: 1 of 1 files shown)


stephengreen (approved):
Looks good, thanks for fixing the bug!

PR #236, opened 2024-02-19
Hotfix for dingo_pipe data_generation

nihargupte-ph:
I am not sure when this bug got introduced, but with the current and previous versions of bilby_pipe, we need to specify args.gaussian_noise and not self.gaussian_noise to be False. Otherwise, this will trigger an error in the data_generation. Namely, see the following line of the bilby source code which is called after running self.create_data https://github.com/lscsoft/bilby_pipe/blob/master/bilby_pipe/data_generation.py#L211.

### DIFF
--- a/dingo/pipe/data_generation.py
+++ b/dingo/pipe/data_generation.py
@@ -156,8 +156,8 @@ class DataGenerationInput(BilbyDataGenerationInput):
             args.injection_dict = None
             args.injection_waveform_arguments = None
             args.injection_frequency_domain_source_model = None
+            args.gaussian_noise = False
             self.frequency_domain_source_model = None
-            self.gaussian_noise = False
 
             self.create_data(args)
 


stephengreen:
Hm, I don't see the offending line in the current version of `bilby_pipe`, see `create_data()` here:

https://git.ligo.org/lscsoft/bilby_pipe/-/blob/master/bilby_pipe/data_generation.py#L211.

The github version of `bilby_pipe` that you linked is three years old. In our `pyproject.toml` we only support `bilby_pipe>=1.3`; are you running an old version?

nihargupte-ph:
> Hm, I don't see the offending line in the current version of `bilby_pipe`, see `create_data()` here:
> 
> https://git.ligo.org/lscsoft/bilby_pipe/-/blob/master/bilby_pipe/data_generation.py#L211.
> 
> The github version of `bilby_pipe` that you linked is three years old. In our `pyproject.toml` we only support `bilby_pipe>=1.3`; are you running an old version?

You are right, in this case I don't think this PR makes sense. Closing it!

PR #240, opened 2024-03-13
Allow asimov to automatically detect the template config file

transientlunatic:
This MR allows bilby to automatically detect the config file template for asimov inside the python package so that it does not need to be specified as a separate template in the `asimov.conf` file.

### DIFF
--- a/dingo/asimov/asimov.py
+++ b/dingo/asimov/asimov.py
@@ -26,7 +26,7 @@ class Dingo(Pipeline):
         Defaults to "C01_offline".
     """
 
-    with importlib.resources.path("dingo.asimov", "datafind_template.yml") as template_file:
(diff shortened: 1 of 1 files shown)


dingo/asimov/asimov.py
```diff
@@ -26,7 +26,9 @@ class Dingo(Pipeline):
         Defaults to "C01_offline".
     """
 
-    # config_template = importlib.resources.path(__name__, 'dingo.ini')
+    with importlib.resources.path("dingo.asimov", "datafind_template.yml") as template_file:
```
nihargupte-ph, line 29:
```suggestion
    with importlib.resources.path("dingo.asimov", "datafind_template.ini") as template_file:
```

nihargupte-ph (approved):

PR #241, opened 2024-04-05
Pytest fixes for inclusion into IGWN

nihargupte-ph:
There are failing pytests configurations which is making it so that we cannot include dingo 0.5.9 into IGWN test. This PR aims to fix the issue of the failing tests (see here: https://git.ligo.org/computing/conda/-/jobs/3268096#L1696). 

There are two failing tests currently, I am attaching the .xml of the failing tests here:
[junit-dingo-gw.txt](https://github.com/dingo-gw/dingo/files/14883697/junit-dingo-gw.txt)

The first test (`test_FD_time_translation_torch`) fails I think because the tolerance of the test is too high. It is time shifting a unit signal by 2pi and checking that it is still a unit signal. I'm not sure how the tolerance was decided. It seems like there are errors of

```
errors = torch.abs(result[..., 0, :] - torch.tensor(1.0))
errors.max() 
>> 3.0e-05
```
This happens when one sets `torch.manual_seed(0)` but doesn't happen for all seeds.

I plot also the histogram of errors here: 
hist_of_errors

Still investigating why the second test fails.

### DIFF
--- a/tests/gw/test_domains.py
+++ b/tests/gw/test_domains.py
@@ -160,7 +160,7 @@ def test_FD_time_translation_torch(uniform_FD_params):
     result = domain.time_translate_data(data, -dt)
     assert result.dtype == torch.float32
     assert result.shape == data.shape
-    assert torch.allclose(result[..., 0, :], torch.tensor(1.0), atol=1e-5)
+    assert torch.allclose(result[..., 0, :], torch.tensor(1.0), atol=1e-4)
     # Tolerance of 1e-2 is required in the imaginary part, likely because we are
     # checking that it *vanishes*. This is a consequence of single-precision floats.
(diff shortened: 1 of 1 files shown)


stephengreen:
I replaced the caching test with one that checks that the updated frequency grid is smaller than the previous one after reducing `_f_max` and resetting the cache.

I'm not sure how the previous test ever passed, since it compares elementwise two arrays of different length. Thoughts @max-dax ?

stephengreen (approved):
Updated tests now pass on my machine, although I am concerned that the previous ones passed at some point as well.

nihargupte-ph:
I tested on CalTech cluster and they seem to pass there too now

PR #245, opened 2024-05-29
Fixing import of lalsimulation.gwsignal to allow for debuggers

nihargupte-ph:
For some reason, when stepping through `import lalsimulation.gwsignal`, the debugger hangs and is unable to step over the import. This is despite when disconnecting the debugger (or alternatively just running the file with regular python) the import works fine. 

* This fix moves the location of the import so that `lalsimulation.gwsignal` is only imported when the `NewInterfaceWaveformGenerator` is instantiated. The functions from `lalsimulation.gwsignal` are then sent to the global namespace. 

* This PR also removes pyseobnr as a required dependency for installation. You can install dingo with pyseobnr by running pip install dingo-gw[pyseobnr]. 

With conda my suggestion would be to not ship pyseobnr with dingo by default and just tell the user they need to install it on their own if they want to use v5.

### DIFF
--- a/dingo/gw/waveform_generator/waveform_generator.py
+++ b/dingo/gw/waveform_generator/waveform_generator.py
@@ -12,9 +12,6 @@ import pandas as pd
 import lal
 import lalsimulation as LS
 
-from lalsimulation.gwsignal.core import waveform
-from lalsimulation.gwsignal.models import gwsignal_get_waveform_generator
-
 from bilby.gw.conversion import (
     convert_to_lal_binary_black_hole_parameters,
     bilby_to_lalsimulation_spins,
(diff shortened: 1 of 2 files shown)


dingo/gw/waveform_generator/waveform_generator.py
```diff
@@ -776,10 +769,25 @@ class NewInterfaceWaveformGenerator(WaveformGenerator):
...
+            # Thus we only import it when the class is instantiated.
+            # However, we don't want to reimport the module every time we need to 
+            # use the function as this is costly. Therefore, we import it once 
+            # when the class is instantiated and store it in the global namespace
+            from lalsimulation.gwsignal.core import waveform
+            from lalsimulation.gwsignal.models import gwsignal_get_waveform_generator
+            globals()['gws_wfm'] = waveform
+            globals()["new_interface_get_waveform_generator"] = gwsignal_get_waveform_generator 
+        except ImportError:
+            raise ImportError("The new interface is not available, please install lalsimulation")
```
stephengreen, line 788: Shouldn't this say to install `pyseobnr`, not `lalsimulation`?

We should make this error message a bit more detailed.

stephengreen:
@nihargupte-ph I think this is going to cause some tests involving SEOBNRv5PHM to fail if `pyseobnr` is not installed:

https://github.com/dingo-gw/dingo/blob/b93eb0367fac2cc1ff948625fe4666fa55d204f0/tests/gw/waveform_generator/test_wfg_m.py#L40

nihargupte-ph:
> @nihargupte-ph I think this is going to cause some tests involving SEOBNRv5PHM to fail if `pyseobnr` is not installed:
> 
> https://github.com/dingo-gw/dingo/blob/b93eb0367fac2cc1ff948625fe4666fa55d204f0/tests/gw/waveform_generator/test_wfg_m.py#L40

Hmm yes you are right. What should be the expected behavior here, I think we should run the pipeline tests with pyseob installed. But for users that don't have pyseob installed maybe we can add a line that checks if pyseob is installed and if it is not it just passes the test?

stephengreen:
I think if pyseobnr is not installed, then skip the test for SEOBNRv5.

PR #246, opened 2024-05-30
FIxing dingo_asd_generation

nihargupte-ph:
This PR fixes condor_submission for generating ASD datasets

*Renamed dingo_merge_ASD_datasets to dingo_merge_asd_datasets in dag creation


annalena-k (approved):
Thanks for the fix!

PR #248, opened 2024-06-11
Fix dingo pipe simple submission

annalena-k:
Since the bilby_pipe update, dingo_pipe was automatically adding the line `ENV GET HTGETTOKENOPTS` in the main `dag_GWxxxxxx.submit`. This line is part of the LIGO authentication ([see LIGO documentation](https://lscsoft.docs.ligo.org/bilby_pipe/v1.3.0/data_acquisition.html)), resulting in an error for `simple-submission=True`.

This PR removes the line for simple submission.

### DIFF
--- a/dingo/pipe/utils.py
+++ b/dingo/pipe/utils.py
@@ -3,6 +3,6 @@ def _strip_unwanted_submission_keys(job):
     job.universe = None
     extra_lines = []
     for line in job.extra_lines:
-        if not line.startswith("priority") and not line.startswith("accounting_group"):
+        if not line.startswith("priority") and not line.startswith("accounting_group") and not line.startswith("ENV"):
             extra_lines.append(line)
     job.extra_lines = extra_lines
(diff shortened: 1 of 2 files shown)


dingo/pipe/utils.py
```diff
@@ -3,6 +3,6 @@ def _strip_unwanted_submission_keys(job):
     job.universe = None
     extra_lines = []
     for line in job.extra_lines:
-        if not line.startswith("priority") and not line.startswith("accounting_group"):
+        if not line.startswith("priority") and not line.startswith("accounting_group") and not line.startswith("ENV"):
```
stephengreen, line 6: Would it work to make this more specific?

```suggestion
        if not line.startswith("priority") and not line.startswith("accounting_group") and not line.startswith("ENV GET HTGETTOKENOPTS"):
```

stephengreen (approved):

dingo/pipe/utils.py
annalena-k: Yes, it works, I just tested it.

PR #251, opened 2024-06-13
Update `result.plot_corner()`

annalena-k:
This PR updates and adds some functionality of `result.plot_corner()`.

It addresses the following problems:
- Previously, the font size of the legend was fixed to 50 which was sufficient for a full 14 dimensional posterior plot. However, if only a few parameters are included via `parameters = ["chirp_mass", "mass_ratio"]`, the legend was so large that it overlapped completely with the corner plot.
- It was not possible to pass the truth values to the corner plot which is helpful for injections.
- So far, the `DINGO-IS` samples were not plotted when `weights` was not included in the parameters list. If the user wanted to only use e.g. `parameters = ["chirp_mass", "mass_ratio"]`, it would only plot the `DINGO` samples.

After this PR, the user can:
- specify the font size of the legend via `legend_font_size=...` and adjust it for a lower dimensional corner plot.
- visualize truth values via `truths=...` as well as the color `truth_color=...`.
- plot the `DINGO-IS` samples although weights are not specified as parameters.

Illustration of an example before the PR where the importance weighted samples are not shown due to the bug:
image

How the plot looks after the PR (when specifying an appropriate font size for the legend):
image

### DIFF
--- a/dingo/core/utils/plotting.py
+++ b/dingo/core/utils/plotting.py
@@ -8,7 +8,12 @@ import pandas as pd
 
 
 def plot_corner_multi(
-    samples, weights=None, labels=None, filename="corner.pdf", **kwargs
+    samples,
+    weights=None,
+    labels=None,
+    filename: str = "corner.pdf",
+    legend_font_size: int = 50,
+    **kwargs,
 ):
     """
     Generate a corner plot for multiple posteriors.
@@ -24,6 +29,8 @@ def plot_corner_multi(
         Labels for the posteriors.
     filename : str
         Where to save samples.
+    legend_font_size: int
+        Font size used in legend. Defaults to 50.
     **kwargs :
         Forwarded to corner.corner.
     """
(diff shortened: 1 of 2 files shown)


nihargupte-ph (approved):
Looks good to me, I tested it locally as well. I changed how the arguments are passed to the functions in the lastest commit. If that is fine then you can merge into main.

dingo/core/utils/plotting.py
```diff
@@ -38,6 +48,8 @@ def plot_corner_multi(
         "levels": [0.5, 0.9],
         "bins": 30,
     }
+    if "truths" in kwargs and kwargs["truths"] is not None and "truth_color" not in kwargs:
```
stephengreen, line 51: I think it's better to just forward all of the settings related to truth values to `corner.corner`, rather than trying to tweak the appearance. It keeps the Dingo code more streamlined. It's okay to use the corner defaults and not include "Truth" in the legend.

PR #252, opened 2024-06-17
Add fixed parameters in the GWSampler

stephengreen:
This adds the parameters that have delta-function priors when sampling. It is needed so that the sky rotation transform and importance sampling works when holding some parameters fixed.

Corner plotting is also updated with a new flag for whether to include these fixed parameters. Default is to not include them.

**This still needs to be tested.**

### DIFF
--- a/dingo/core/result.py
+++ b/dingo/core/result.py
@@ -589,6 +589,7 @@ class Result(DingoDataset):
         parameters: list = None,
         filename: str = "corner.pdf",
         truths: dict = None,
+        include_fixed_parameters: bool = False,
         **kwargs,
     ):
         """
@@ -603,6 +604,8 @@ class Result(DingoDataset):
             Where to save samples.
         truths : dict
             Dictionary of truth values to include.
+        include_fixed_parameters : bool
+            Whether to plot parameters that have delta-function priors. (Default: False)
 
         Other Parameters
         ----------------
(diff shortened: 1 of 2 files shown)


annalena-k:
Yes, this PR fixes running injections with fixed parameters.

I tested the PR on the tutorial example and the following code worked to generate an injection with fixed posterior parameters except for chirp mass and mass ratio:

```
    # Load the network
    pm = PosteriorModel(model_path, device="cpu")
    sampler = GWSampler(model=pm)

    # Generate an injection consistent with the data the model was trained on.
    injection = Injection.from_posterior_model_metadata(pm.metadata)
    injection.asd = ASDDataset(asd_path, ifos=["H1", "L1"])
    theta = injection.prior.sample()
    inj = injection.injection(theta)

    # Generate 10,000 samples from the normalizing flow.
    sampler.context = inj
    sampler.run_sampler(10000)

    result = sampler.to_result()

    # Run importance-sampling on the result.
    num_processes = 8
    result.importance_sample(num_processes=num_processes)

    # Make a corner plot and save the result.
    result.print_summary()
```

However, running `dingo_pipe` on a model with Delta priors does not work (the weights are nan?). I will do some more investigation to track down the bug.

max-dax (approved):

stephengreen:
Try it now. I'm excluding the fixed parameters when evaluating the prior.

annalena-k:
I tested the new commit on dingo_pipe of the tutorial example with chirp mass and mass ratio:
- It works and runs through.
- The corner plot contains chirp mass, mass ratio, log prob, log prior, log likelihood and weights (as expected).

Example plot:
image

This should be sufficient for the tutorial. You can merge the branch.

annalena-k (approved):
Seems to work now for the tutorial as intended. Thanks!

PR #253, opened 2024-06-18
Update plot corner (additional changes)

annalena-k:
Upon discussion with Stephen, I changed how the truth values and truth color are passed to `result.plot_corner()` and `plot_corner_multi()`.

The reasoning behind this is the following:
- `truth_color` is included in the `**kwargs` and the description is removed since it is exactly a `corner` kwarg and does not get modified.
- `truths` is passed as a kwarg to `plot_corner_multi()`. There, `truths` needs to be a list with the values in exactly the same order as the values within `samples`. However, this order is specified within `result.plot_corner()` by the order of the `parameter` list. Therefore, we need to pass the truth values as a dictionary to `result.plot_corner()` and extract the order of parameters by replicating the order of `theta.columns`. The resulting list is added to kwargs and passed on to `plot_corner_multi()`.

I am not entirely sure whether this is the best way of handling the problem since it might be confusing why `truth_color` is handled as a kwarg in `result.plot_corner()` while `truths` is an explicit argument.

We can discuss this within this PR.

### DIFF
--- a/dingo/core/result.py
+++ b/dingo/core/result.py
@@ -588,9 +588,8 @@ class Result(DingoDataset):
         self,
         parameters: list = None,
         filename: str = "corner.pdf",
-        legend_font_size: int = 50,
         truths: dict = None,
-        truth_color: str = None,
+        **kwargs,
     ):
         """
         Generate a corner plot of the samples.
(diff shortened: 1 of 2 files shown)


dingo/core/result.py
```diff
@@ -595,27 +601,43 @@ def plot_corner(self, parameters=None, filename="corner.pdf"):
...
 
+        if "weights" in theta:
+            weights = theta["weights"]
+        else:
+            weights = None
         # User option to plot specific parameters.
         if parameters:
             theta = theta[parameters]
+            if truths is not None:
+                kwargs["truths"] = [truths.get(k) for k in theta.columns]
```
stephengreen, line 625: Should this not be outside the previous `if`?
```suggestion
        if truths is not None:
            kwargs["truths"] = [truths.get(k) for k in theta.columns]
```

annalena-k:
I checked the functionality again, it works as intended:
- for `include_fixed_params=True` corner throws an expected error (no dynamic range)
- if you don't specify anything, `result.plot_corner` includes the non-fixed parameters as well as log_prob, log_prior, weights, etc.
- if you specify `parameters=["chirp_mass", "mass_ratio"]`,  you only get the 2D corner plot.

stephengreen (approved):

PR #258, opened 2024-10-08
Adding support for SEOBNRv4EHM, SEOBNRv4HM, NRSur7dq4 networks

nihargupte-ph:
This PR adds support for SEOBNRv4EHM, SEOBNRv4HM and NRSur7dq4. In particular it implements TD conditions for waveform modes so that we can use the synthetic phase. 

I removed most things not related to waveform generation from the PR to keep it clean (dingo_pipe injections which still needs to get merged)


nihargupte-ph:
Deprecating this, as SEOBNRv5+ is the model planned to be used for future runs.

PR #260, opened 2024-10-15
Removed bid argument from examples yaml files

nihargupte-ph:
The bid argument is only useful for the MPI-IS cluster so removing it from the examples.


annalena-k (approved):
Looks good!
Thanks Nihar!

PR #263, opened 2024-10-18
Imr phenom xhm support

sclyne623:
adds the IMRPhenomXHM approximant to the list of valid approximants in the `generate_FD_modes_LO function` and added a conditional check before `convert_J_to_L0_frame` as IMRPhenomXHM is not generated in the J frame.

### DIFF
--- a/dingo/gw/waveform_generator/waveform_generator.py
+++ b/dingo/gw/waveform_generator/waveform_generator.py
@@ -660,7 +660,8 @@ class WaveformGenerator:
         """
         # TD approximants that are implemented in J frame. Currently tested for:
         #   101: IMRPhenomXPHM
-        if self.approximant in [101]:
+        tested_approximants = ["IMRPhenomXPHM","IMRPhenomXHM"]
+        if LS.GetStringFromApproximant(self.approximant) in tested_approximants:
             parameters_lal_fd_modes = self._convert_parameters(
                 {**parameters, "f_ref": self.f_ref},
                 lal_target_function="SimInspiralChooseFDModes",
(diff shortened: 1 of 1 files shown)


mpuerrer (commented):
It would be good to add a comment which explains why IMRPhenomXHM behaves differently from IMRPhenomXPHM. In the code we see that both approximants are supported in the if statement in line 664, but in line 679 the check is only for IMRPhenomXPHM. Don't leave the reader guessing, but give a clear and concise statement about why they need to be treated differently.

mpuerrer (commented):
You deleted the comment about "approximants that are implemented in J frame"; when you submit a merge request like this one, it is always good to add a paragraph motivating why the changes are necessary and what specific problem they fix. Please also add this reasoning for this merge request. in the conversation tab.

mpuerrer (commented):
Thanks for updating line 661. If you read through the comments in the entire function, you'll see that there are other places where the comments should be updated: in line 674 it would be good to specify that "For some waveform models considered here ... the modes are returned in the J-frame".

And it would be good to add near the beginning of the function -- line 662 which frame each of the waveform model's modes are give in, perhaps like this:
```
# IMRPhenomXPHM: modes in J-frame
# IMRPhenomXHM: modes in L-frame
```

It also isn't necessary to list the numbers 101, 97 anymore because we have the names and the numbers come from a C enum in LALSimulation which would change if someone were to insert a waveform before a particular approximant.

mpuerrer (commented):
Thanks, this looks good to me.

mpuerrer:
@stephengreen please have a look as well if you are happy with this change.

stephengreen (changes requested):
Thanks @sclyne623. To be sure that this works properly, please extend the test https://github.com/dingo-gw/dingo/blob/7fd8ac8c355e5864ca9dde2dbc80d74d5d567cb2/tests/gw/waveform_generator/test_wfg_m.py#L142
to include IMRPhenomXHM.

PR #264, opened 2024-10-22
FEAT: add argument to specify calibration correction direction

nihargupte-ph:
Following this bilby PR https://github.com/bilby-dev/bilby/pull/47 we can update the ApplyCalibrationUncertainty transform and dingo_pipe whether the calibration envelope is on $\eta$ or $\alpha = 1 / \eta$. Here $d = \alpha h + n$.

### DIFF
--- a/dingo/gw/transforms/detector_transforms.py
+++ b/dingo/gw/transforms/detector_transforms.py
@@ -254,7 +254,7 @@ class ApplyCalibrationUncertainty(object):
     is not perfectly calibrated, the observed waveform is not identical to the true
     waveform $h(f)$. Rather, it is assumed to have corrections of the form
 
-    $$h_{obs}(f) = h(f) * (1 + \delta A(f)) * \exp(i \delta \phi(f)),$$
+    $$h_{obs}(f) = h(f) * (1 + \delta A(f)) * \exp(i \delta \phi(f)) = h(f) * \alpha(f),$$
 
     where $\delta A(f)$ and $\delta \phi(f)$ are frequency-dependent amplitude and
     phase errors. Under the calibration model, these are parametrized with cubic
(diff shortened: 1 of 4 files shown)


dingo/gw/transforms/detector_transforms.py
```diff
@@ -340,6 +356,7 @@ def __init__(
                     self.data_domain.f_max,
                     num_calibration_nodes,
                     ifo.name,
+                    correction_type=correction_type,
```
stephengreen, line 359: So `correction_type` just gets forwarded to the relevant Bilby object? Are the terms "data" and "template" taken from there?

Do we need to specify a minimum Bilby version for this to work?

dingo/pipe/parser.py
```diff
@@ -99,6 +99,13 @@ def create_parser(top_level=True):
         help=("Number of calibration curves to use in marginalizing over calibration "
               "uncertainty"),
     )
+    calibration_parser.add(
+        "--calibration-correction-type",
+        type=str,
+        default="data",
+        help=("Number of calibration curves to use in marginalizing over calibration "
```
stephengreen, line 106: Is this taken directly from the updated `bilby_pipe`? The help string seems wrong.

Do we need to update the `bilby_pipe` version?

stephengreen (changes requested):

dingo/gw/transforms/detector_transforms.py
nihargupte-ph: I don't think we need to remove support for older versions of Bilby just for this though. Many users won't even use calibration marginalization so this won't affect them.

dingo/pipe/parser.py
ColmTalbot: We haven't pushed this feature to bilby_pipe yet, that will happen after we merge into bilby.

nihargupte-ph: I just updated this to match the bilby_pipe helpstring, so it should be good to go. Also changed requirement to bilby_pipe == 1.4.2

max-dax (approved):

PR #268, opened 2024-11-18
authetication changes and change bilby version to 1.4.2

nihargupte-ph:


stephengreen (approved):
Looks good, I just adjusted the positioning of the scitoken-issuer within the parser file, for consistency with bilby-pipe.

PR #274, opened 2025-01-07
Fix bugs related to early stopping and wandb

annalena-k:
This PR fixes two small bugs:
1. The argument `metric` of `EarlyStopping` does not have an effect at the moment since it is not used in the code after [initialization](https://github.com/dingo-gw/dingo-development/blob/204e43359d74fa817cf17a657c54bc52e20e5451/dingo/core/utils/trainutils.py#L55). Since it is mentioned in the example train settings files [(for example here)](https://github.com/dingo-gw/dingo/blob/11c8269aa8a76fcd51aad430822d8c8edc91d1e3/examples/fmpe_model/train_settings.yaml#L96), this might give a wrong impression to users.
A simple check for `early_stopping.metric` [here](https://github.com/dingo-gw/dingo/commit/ca79cbde9e59187f18a29db223dc411a2ef117b8#diff-e4df0a069e94766d19de6ff0aa66ff79d349e278b7a519d696e7c4c715df308bR455) which determines whether the training or test loss is used to check for early stopping resolves this issue.
2. `wandb` can crash the training on an htcondor cluster, because `$HOME` is often not automatically defined in a submitted job. This is detrimental because the `.netrc` file (which specifies the `wandb` user credentials) is located there. Without the user credentials, `wandb` can't log correctly and it crashes the job.
This is solved by adding the line `getenv = True` to the submission file.

### DIFF
--- a/dingo/core/posterior_models/base_model.py
+++ b/dingo/core/posterior_models/base_model.py
@@ -2,6 +2,7 @@
 This module contains the abstract base class for representing posterior models,
 as well as functions for training and testing across an epoch.
 """
+
 from abc import abstractmethod, ABC
 import os
 from os.path import join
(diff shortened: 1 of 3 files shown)


nihargupte-ph (approved):
LGTM

PR #276, opened 2025-01-20
asimov updates to handle calibration and bugfix with data folder and …

nihargupte-ph:
The asimov ini file was not correctly taking the Cubicspline argument if it was specified. The changes to `dingo.ini` fix this. 

Also, because we have a folder named `data` in the DINGO run directory, ln 281 of `dingo/pipe/main.py` was changing the calibration correction type argument (which could be `data`) to `/abspath_of_dingo_dir/data` which causes the calibration envelope to point to the wrong location.


dingo/pipe/main.py
```diff
@@ -235,6 +235,7 @@ def __init__(self, args, unknown_args, importance_sampling_updates):
         self.psd_fractional_overlap = args.psd_fractional_overlap
         self.psd_start_time = args.psd_start_time
         self.spline_calibration_envelope_dict = args.spline_calibration_envelope_dict
+        self.calibration_correction_type = args.calibration_correction_type
```
stephengreen, line 238: Can you remind me why we have to specify this calibration information in `MainInput`? In bilby-pipe, it's only done at a later stage. I'm sure we did it for a reason, I just can't remember.

stephengreen (approved):
Looks good to me. Just one question in the code.

dingo/pipe/main.py
nihargupte-ph: I'm trying to remember now and can't... I will try to test what happens

nihargupte-ph: Ok, so this is a good point. It's better to do this where the data generation happens. That way you don't have to also modify the main Input. I tested locally w/ the debugger and seems to propagate the calibration arguments.

Issue #277, opened 2025-01-20
Bug in FMPE training


annalena-k:
When running FMPE training with `posterior_model_type: flow_matching`, I get the following error:

```
Beginning new training run.
Loading dataset from /Users/akofler/Desktop/gravitational_waves/code_production/dingo_tutorial/01_training_data/waveform_dataset/waveform_dataset.hdf5.

Building SVD for initialization of embedding network.
Setting train transforms.
Omitting 
	AddWhiteNoiseComplex
	RepackageStrainsAndASDS
	SelectStandardizeRepackageParameters
	UnpackDict
Loading dataset from /Users/akofler/Desktop/gravitational_waves/code_production/dingo_tutorial/01_training_data/asd_dataset/asd_dataset_GW150914.hdf5.
Calculating new parameter standardizations.
Generating waveforms for embedding network SVD initialization.
...done. This took 1 s.
Generating SVD basis for ifo:
...H1 done.
...L1 done.
...this took 10 s.
Testing SVD basis matrices.
...H1:
n = 50
  Mean mismatch = 1.4157657823865577e-11
  Standard deviation = 1.165590464587736e-11
  Max mismatch = 7.944533919612695e-11
  Median mismatch = 1.0669742867008836e-11
  Percentiles:
    99    -> 6.385663220420668e-11
    99.9  -> 7.765800391634786e-11
    99.99 -> 7.926660566814894e-11
n = 100
  Mean mismatch = 6.413296560481285e-13
  Standard deviation = 4.733723277443611e-13
  Max mismatch = 4.463207581295592e-12
  Median mismatch = 5.144773496112975e-13
  Percentiles:
    99    -> 2.3527579884330398e-12
    99.9  -> 3.888798061346729e-12
    99.99 -> 4.4057666293006725e-12
Saving dataset to dingo_tutorial/02_training_fmpe/svd_H1.hdf5
...L1:
n = 50
  Mean mismatch = 1.4167102269091458e-11
  Standard deviation = 9.912647112632404e-12
  Max mismatch = 6.429212717762312e-11
  Median mismatch = 1.1604439631440755e-11
  Percentiles:
    99    -> 5.139694003730709e-11
    99.9  -> 6.107514549658106e-11
    99.99 -> 6.397042900951872e-11
n = 100
  Mean mismatch = 7.763654163994715e-13
  Standard deviation = 6.266360265319222e-13
  Max mismatch = 4.765743355505947e-12
[...]

PR #278, opened 2025-01-20
Fix fmpe bug related to the cache

annalena-k:
Fix for this issue: https://github.com/dingo-gw/dingo/issues/277

I verified that training is now running successfully with a toy example.


jowi-calliora (approved):

jonaswildberger (approved):
LGTM. Good catch. Thank you!

Issue #283, opened 2025-02-05
Unconditional density estimation is broken due to Identity


nihargupte-ph:
When training an unconditional density estimator one does not pass context. Currently, the embedding net is set to an identity for the unconditional density estimator. However, this still requires context which causes an error (see below):

```
    unconditional_model = train_unconditional_density_estimator(
  File "/work/nihargupte/src/dingo-unstable/dingo/core/density/unconditional_density_estimation.py", line 103, in train_unconditional_density_estimator
    model.train(
  File "/work/nihargupte/src/dingo-unstable/dingo/core/posterior_models/base_model.py", line 407, in train
    train_loss = train_epoch(self, train_loader)
  File "/work/nihargupte/src/dingo-unstable/dingo/core/posterior_models/base_model.py", line 481, in train_epoch
    loss = pm.loss(data[0], *data[1:])
  File "/work/nihargupte/src/dingo-unstable/dingo/core/posterior_models/normalizing_flow.py", line 62, in loss
    return -self.network(theta, *context).mean()
  File "/work/nihargupte/environments/micromamba/envs/dingo-asimov/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/work/nihargupte/environments/micromamba/envs/dingo-asimov/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
  File "/work/nihargupte/src/dingo-unstable/dingo/core/nn/nsf.py", line 241, in forward
    return self.log_prob(y)
  File "/work/nihargupte/src/dingo-unstable/dingo/core/nn/nsf.py", line 211, in log_prob
    x = self.embedding_net(*x)
  File "/work/nihargupte/environments/micromamba/envs/dingo-asimov/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/work/nihargupte/environments/micromamba/envs/dingo-asimov/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
TypeError: forward() missing 1 required positional argument: 'input'    unconditional_model = train_unconditional_density_estimator(
[...]

PR #284, opened 2025-02-05
Prevent FlowWrapper from using embedding net when there is no context.  

nihargupte-ph:
This PR fixes this bug: https://github.com/dingo-gw/dingo/issues/283 which currently breaks the unconditional density estimation. It only does so by passing context to the embedding net when there is context.

### DIFF
--- a/dingo/core/nn/nsf.py
+++ b/dingo/core/nn/nsf.py
@@ -207,27 +207,27 @@ class FlowWrapper(nn.Module):
         self.flow = flow
 
     def log_prob(self, y, *x):
-        if self.embedding_net is not None:
-            x = self.embedding_net(*x)
         if len(x) > 0:
+            if self.embedding_net is not None:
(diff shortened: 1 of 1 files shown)


stephengreen:
I think we also want to resolve here whether we need default `nn.Identity()` embedding nets, as sometimes occur. It would seem to me that this is undesirable, but maybe there is a good reason.

stephengreen:
I've removed the `nn.Identity()` in `FlowWrapper`. I tested training and inference for an NPE model and had no problems.

We were also storing the model hyperparameters as an attribute of the Flow, but as far as I could tell they were never accessed. I removed this line as well.

@nihargupte-ph Please test and if you are happy I will approve.

stephengreen (approved):
Please test this and then merge. I made a few changes.

nihargupte-ph:
I tested inference, seems to work (22% efficiency in GW150914)

PR #285, opened 2025-02-12
using torch loading with fallback

vincentberenz:
## About

The toy npe demo could not run using older version of python and cuda (e.g. python3.9 / cuda 11.3), because of poor support of the "meta" device. This pull request introduces the function 'torch_load_with_fallback' which fallback to other devices when the preferred device is not available/supported. This function attempts to fallback to cuda/mps/hip/cpu 

The code has been updated to use `torch_load_with_fallback` everywhere, not only to solve the compatibility with older version, but also for more flexibility in general.

## How we know it works

- pytest passing (cpu / older cuda)
- toy npe demo passing (cpu/older cuda)

Not tested for mps/hip.

## To consider

- Some modules in the sub-package `compatibility` where independent to the dingo package. Using `torch_load_with_fallback` them made them dependent to it. This should be reverted if not a good thing.

### DIFF
--- a/dingo/core/posterior_models/base_model.py
+++ b/dingo/core/posterior_models/base_model.py
@@ -3,26 +3,24 @@ This module contains the abstract base class for representing posterior models,
 as well as functions for training and testing across an epoch.
 """
 
-import json
+from abc import abstractmethod, ABC
 import os
-import time
-from abc import ABC, abstractmethod
-from collections import OrderedDict
 from os.path import join
-from typing import Optional
-
 import h5py
-import numpy as np
-import torch
-from threadpoolctl import threadpool_limits
-from torch.utils.data import Dataset
 
+import torch
 import dingo.core.utils as utils
+from torch.utils.data import Dataset
+import time
+import numpy as np
+from threadpoolctl import threadpool_limits
 import dingo.core.utils.trainutils
-from dingo.core.utils.backward_compatibility import (
-    update_model_config,
-)
+import json
+from collections import OrderedDict
+from typing import Optional
+from dingo.core.utils.backward_compatibility import update_model_config
 from dingo.core.utils.misc import get_version
+
 from dingo.core.utils.trainutils import EarlyStopping
 
 
@@ -83,9 +81,7 @@ class BasePosteriorModel(ABC):
         # build model
         if model_filename is not None:
             self.load_model(
-                model_filename,
-                load_training_info=load_training_info,
-                device=device,
+                model_filename, load_training_info=load_training_info, device=device
             )
         else:
             self.initialize_network()
@@ -121,9 +117,7 @@ class BasePosteriorModel(ABC):
         pass
 
     @abstractmethod
-    def sample_and_log_prob(
-        self, *context: torch.Tensor, num_samples: int = 1
-    ):
+    def sample_and_log_prob(self, *context: torch.Tensor, num_samples: int = 1):
         """
         Sample parameters theta from the posterior model,
 
@@ -196,9 +190,7 @@ class BasePosteriorModel(ABC):
         Put model to device, and set self.device accordingly.
         """
         if device not in ("cpu", "cuda"):
-            raise ValueError(
-                f"Device should be either cpu or cuda, got {device}."
-            )
+            raise ValueError(f"Device should be either cpu or cuda, got {device}.")
         self.device = torch.device(device)
         # Commented below so that code runs on first cuda device in the case of multiple.
         # if device == 'cuda' and torch.cuda.device_count() > 1:
@@ -260,13 +252,9 @@ class BasePosteriorModel(ABC):
             model_dict["optimizer_kwargs"] = self.optimizer_kwargs
             model_dict["scheduler_kwargs"] = self.scheduler_kwargs
             if self.optimizer is not None:
-                model_dict["optimizer_state_dict"] = (
-                    self.optimizer.state_dict()
-                )
+                model_dict["optimizer_state_dict"] = self.optimizer.state_dict()
             if self.scheduler is not None:
-                model_dict["scheduler_state_dict"] = (
-                    self.scheduler.state_dict()
-                )
+                model_dict["scheduler_state_dict"] = self.scheduler.state_dict()
 
         torch.save(model_dict, model_filename)
 
@@ -301,9 +289,7 @@ class BasePosteriorModel(ABC):
             # Load model weights
             model_state_dict = OrderedDict()
             for k, v in fp["model_weights"].items():
-                model_state_dict[k] = torch.from_numpy(
-                    np.array(v, dtype=np.float32)
-                )
+                model_state_dict[k] = torch.from_numpy(np.array(v, dtype=np.float32))
             d["model_state_dict"] = model_state_dict
 
         return d
@@ -444,12 +430,8 @@ class BasePosteriorModel(ABC):
                 utils.perform_scheduler_step(self.scheduler, test_loss)
 
                 # write history and save model
-                utils.write_history(
-                    train_dir, self.epoch, train_loss, test_loss, lr
-                )
-                utils.save_model(
-                    self, train_dir, checkpoint_epochs=checkpoint_epochs
-                )
+                utils.write_history(train_dir, self.epoch, train_loss, test_loss, lr)
+                utils.save_model(self, train_dir, checkpoint_epochs=checkpoint_epochs)
                 if use_wandb:
                     try:
                         import wandb
@@ -467,9 +449,7 @@ class BasePosteriorModel(ABC):
                             }
                         )
                     except ImportError:
-                        print(
-                            "wandb not installed. Skipping logging to wandb."
-                        )
+                        print("wandb not installed. Skipping logging to wandb.")
 
                 if early_stopping is not None:
                     # Whether to use train or test loss
@@ -481,8 +461,7 @@ class BasePosteriorModel(ABC):
                     is_best_model = early_stopping(early_stopping_loss)
                     if is_best_model:
                         self.save_model(
-                            join(train_dir, "best_model.pt"),
-                            save_training_info=False,
+                            join(train_dir, "best_model.pt"), save_training_info=False
                         )
                     if early_stopping.early_stop:
                         print("Early stopping")
--- a/dingo/core/posterior_models/base_model.py
+++ b/dingo/core/posterior_models/base_model.py
@@ -18,7 +18,7 @@ import dingo.core.utils.trainutils
 import json
 from collections import OrderedDict
 from typing import Optional
-from dingo.core.utils.backward_compatibility import update_model_config
+from dingo.core.utils.backward_compatibility import update_model_config, torch_load_with_fallback
 from dingo.core.utils.misc import get_version
 
 from dingo.core.utils.trainutils import EarlyStopping
@@ -318,7 +318,7 @@ class BasePosteriorModel(ABC):
         # machine may have moved the model from 'cuda' to 'cpu'.
         ext = os.path.splitext(model_filename)[-1]
         if ext == ".pt":
-            d = torch.load(model_filename, map_location=device)
+            d = torch_load_with_fallback(model_filename, preferred_map_location=device)
         elif ext == ".hdf5":
             d = self._load_model_from_hdf5(model_filename)
         else:
--- a/dingo/core/posterior_models/base_model.py
+++ b/dingo/core/posterior_models/base_model.py
@@ -84,7 +84,9 @@ class BasePosteriorModel(ABC):
         # build model
         if model_filename is not None:
             self.load_model(
-                model_filename, load_training_info=load_training_info, device=device
+                model_filename,
+                load_training_info=load_training_info,
+                device=device,
             )
         else:
             self.initialize_network()
@@ -120,7 +122,9 @@ class BasePosteriorModel(ABC):
         pass
 
     @abstractmethod
-    def sample_and_log_prob(self, *context: torch.Tensor, num_samples: int = 1):
+    def sample_and_log_prob(
+        self, *context: torch.Tensor, num_samples: int = 1
+    ):
         """
         Sample parameters theta from the posterior model,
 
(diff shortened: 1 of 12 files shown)


dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,80 @@
...
+) -> Tuple[Dict, Device]:
+    """
+    Loads a PyTorch file with fallback behavior:
+    1. Tries preferred_map_location (default: cuda)
+    2. Falls back to CUDA/MPS/HIP if available
+    3. Finally falls back to CPU
+
+    Returns
+    -------
+    Loaded model
```
annalena-k, line 54: Can you please update the doc-string to clarify that the first return argument is the model and the second is a device?
What is the type of the device? Is it a string or is it a torch.device?

dingo/core/posterior_models/base_model.py
```diff
@@ -318,7 +321,9 @@ def load_model(
         # machine may have moved the model from 'cuda' to 'cpu'.
         ext = os.path.splitext(model_filename)[-1]
         if ext == ".pt":
-            d = torch.load(model_filename, map_location=device)
+            d, _ = torch_load_with_fallback(
```
annalena-k, line 324: I think there is one more philosophical question that we have to discuss regarding this change:
In some locations in the code we don't care to which device the model is loaded but prefer one device (e.g. when we just want to extract the metadata) while the model should not be put on another device than specified in other parts of the code.
For example, here, we want to be sure 100% that the model is put on the provided `preferred_map_location=device` because later we will put the training data to the same `device`. As a result, I would expect an error message when I try to start a training run with a GPU and `device='cuda'`, but no GPU is available. We would also run into issues, when the model is put onto a different device than the data.
Maybe we need different options, e.g. a `preferred_map_location` argument and `map_location` argument where the first is flexible and the second is fixed and results in an error if the device is not available?

I definitely want @stephengreen's opinion on this.

dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,80 @@
...
+def torch_available_devices() -> List[Device]:
+    """
+    Returns a list of all available PyTorch devices,
+    ordered: meta, cuda, mps, hip, cpu
+
+    Returns
+    -------
+    List of available device identifiers
+    """
+    devices: List[Device] = ["meta"]
```
annalena-k, line 17: We should discuss in which order we want to have the fallback options. How it is written at the moment, it will first try to load a model to "meta", then "cuda", etc.
However, "meta" as the first option has the problem that the model parameters are not loaded into memory. If a model is accidentally loaded with "meta", training would fail with a very weird error.
Therefore, "meta" should be the last option (even after cpu), if not specified on purpose.
For the rest of the ordering, I don't have a preference.

annalena-k: An example for why this is not a good order at the moment:
If I specify `device=cuda`, but no GPU is available, the loading will fail and it will try to find a fallback option. The first option here is "meta" which would load the model without parameters and therefore throw a weird error when starting training.
(But there's another comment above discussing this case since we want to see an error. Just not a heard-to-understand meta-state error message)

annalena-k (changes requested):
See comments.

dingo/core/posterior_models/base_model.py
vincentberenz: Supporting both preferred_map_location and map_location as suggested sounds good, I see no downside. We will just need to make sure we have no "hard coded" map_location, e.g. `map_location='cuda'`, which would fail on systems not supporting cuda (but we could do `map_location= as in my configuration file`)

dingo/core/utils/backward_compatibility.py
vincentberenz: what about an argument "fallback_devices" ?
```python
# fallback_devices in order of preference
model, device ='torch_load_with_fallback(path, preferred_device="meta", fallback_devices=("cpu",))
```

dingo/core/posterior_models/base_model.py
annalena-k: But let's wait for Stephen's opinion on this before you make all the changes.

dingo/core/utils/backward_compatibility.py
annalena-k: I think it's a good idea to specify the fallback options in one place (rather than having to think about it every time we load a model).
I would only move "meta" to the last option.

dingo/core/posterior_models/base_model.py
vincentberenz: Sure !
But just realized: when we really want a specific device, we can directly use the original `torch.load` function instead of `torch_load_with_fallback`. I changed all calls from `torch.load` to `torch_load_with_fallback`, but likely some of them should be simply be reverted to `torch.load`.

dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,80 @@
...
+    1. Tries preferred_map_location (default: cuda)
+    2. Falls back to CUDA/MPS/HIP if available
+    3. Finally falls back to CPU
+
+    Returns
+    -------
+    Loaded model
+    """
+
+    try:
```
nihargupte-ph, line 57: Maybe one could add a print statement to indicate which device is being loaded? My suggestion would be: 

```suggestion
devices = torch_available_devices()
devices.insert(0, preferred_map_location)
for location in devices:
   try: 
      print(f"Attempting to load the model on {location}")
      return torch.load(filename, map_location=location), location
   except RuntimeError:
      print(f"Failed to load model on {location}, check your system settings")
      pass
```

That way if the situation Annalena mentioned happens (CPU training) the user would know in the logs and be able to debug from there.

dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,80 @@
...
+    except RuntimeError:
+        pass
+
+    devices = torch_available_devices()
+
+    for location in [d for d in devices if d != preferred_map_location]:
+        try:
+            return torch.load(filename, map_location=location), location
+        except RuntimeError:
+            pass
```
nihargupte-ph, line 71: Maybe one could add a print statement to indicate which device is being loaded? My suggestion would be: 

```suggestion
devices = torch_available_devices()
devices.insert(0, preferred_map_location)
for location in devices:
   try: 
      print(f"Attempting to load the model on {location}")
      return torch.load(filename, map_location=location), location
   except RuntimeError:
      print(f"Failed to load model on {location}, check your system settings")
      pass
```

That way if the situation Annalena mentioned happens (CPU training) the user would know in the logs and be able to debug from there.

nihargupte-ph, line 71:
```suggestion
    devices = torch_available_devices()
    devices.insert(0, preferred_map_location)
    for location in devices:
            try: 
                print(f"Attempting to load the model on {location}")
                return torch.load(filename, map_location=location), location
        
        except RuntimeError:
            print(f"Failed to load model on {location}, check your system settings")
            pass           
```

dingo/core/posterior_models/base_model.py
annalena-k: I think the way to go is the following:
- We want to use the standard `torch.load` when we actually want to load the model with all parameters, because the model has to be put on `device`.
- If we only want to access `pm.metadata`, we want the fallback. (I.e., this should be the case at all location where the code said `device = 'meta'`).

dingo/core/utils/backward_compatibility.py
vincentberenz: I just updated:
- the returns doc is better
- the type of the device was str, it is not torch.device
Note: I do not specify the type in the docstring. This is because the function has typing hints, so the return type is already specified. Once the code fully typed, we can build the documentation so that the typing hints are used during the html generation.

dingo/core/utils/backward_compatibility.py
vincentberenz: Yes, this is indeed slighly more elegant, but I just updated the code to use logging, with a message which is a bit different depending on if the preferred map location could be used for not.

dingo/core/utils/backward_compatibility.py
annalena-k: Ah, interesting!

dingo/core/posterior_models/base_model.py
```diff
@@ -3,24 +3,27 @@
...
+from torch.utils.data import Dataset
+
+import dingo.core.utils as utils
 import dingo.core.utils.trainutils
-import json
-from collections import OrderedDict
-from typing import Optional
-from dingo.core.utils.backward_compatibility import update_model_config
+from dingo.core.utils.backward_compatibility import (
+    torch_load_with_fallback,
```
annalena-k, line 23: I just checked we are not using `torch_load_with_fallback` within `base_model.py`
```suggestion
```

dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,96 @@
...
+
+
+def torch_load_with_fallback(
+    filename: str, preferred_map_location: Device = "cuda"
+) -> Tuple[Dict, torch.device]:
+    """
+    Loads a PyTorch file with fallback behavior:
+    1. Tries preferred_map_location (default: cuda)
+    2. Falls back to CUDA/MPS/HIP if available
+    3. Finally falls back to CPU
```
annalena-k, line 56: Do you want to mention `meta` here as well as the final fallback?

dingo/core/posterior_models/base_model.py
vincentberenz: nice catch !

dingo/core/utils/backward_compatibility.py
vincentberenz: Just added it. (but I think CPU shall never fail ?)

annalena-k: Yeah, true :D

vincentberenz:
@stephengreen Annalena checked everything, shall we merge this ?

annalena-k:
Sorry, one more thing: Can you merge main into this branch to make sure the tests don't fail like in Nihar's case?

dingo/core/posterior_models/base_model.py
stephengreen: I agree with @annalena-k's comments here. In this context, we want to load to a specific requested device, so we should maintain the original code (as seems to be the case in the latest version). In other cases, we don't care so much.

To add to @annalena-k's list, there are some cases where we want to modify the metadata and then save the whole model + metadata (e.g., in the backward compatibility scripts). Here, I don't believe we can use the `meta` device.

dingo/core/utils/backward_compatibility.py
stephengreen: I agree that we want to be careful about loading to "meta", particularly when we want to access the actual parameters. As I mentioned in my other comment, we might want to remove "meta" entirely from the fallback list. In addition, I don't think there is any situation where we would fail to load to "cpu", so we will probably never reach "meta" in the list.

stephengreen (changes requested):
Thanks @vincentberenz! There are just a few comments, a couple of which look like typos. I think we might want to remove "meta" from the fallback list as well.

I was also curious what the situation is where we only have older CUDA? Old hardware?

dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,96 @@
+from typing import Dict, List, Literal, Tuple
+import logging
+
+import torch
+
+_logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
vincentberenz: Using logging, the setup can be done at the beginning of the executable. This consists mainly at setup the level of what will be printed (e.g. if level is info, the debug level is ignored) and what will be the stdout(s) (e.g. the terminal, or to files, or both).  If the executable takes arguments, we can add arguments for the setup of the log, so that each user can print what/where they want (e.g. `python myscript --debug`).   

"Will this just print a statement to the standard output?". I think the default setup is printing level info in the terminal.

I propose we start moving from prints to logging. We could make one big pull request, or step by step, as I started to do here.  Downside is that we would be hybrid prints/logging for a while. I do not have strong opinion.

In any case, I am using logging for the dingo waveform refactor.

dingo/gw/ls_cli.py
```diff
@@ -1,17 +1,18 @@
...
 from pathlib import Path
+from pprint import pprint
 
 import h5py
 import torch
 import yaml
-import json
-from pprint import pprint
 
+from dingo.core.backward_compatibility import torch_load_with_fallback
```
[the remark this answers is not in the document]
vincentberenz: indeed !

dingo/core/posterior_models/base_model.py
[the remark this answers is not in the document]
vincentberenz: No, I restored the non black formatted file for clarity

dingo/core/utils/backward_compatibility.py
```diff
@@ -1,3 +1,96 @@
...
+
+Device = Literal["meta", "cuda", "mps", "hip", "cpu"]
+
+
+def torch_available_devices() -> List[Device]:
+    """
+    Returns a list of all available PyTorch devices,
+    ordered: cuda, mps, hip, cpu, meta
```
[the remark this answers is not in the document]
vincentberenz: Indeed, I removed meta from the list

dingo/core/posterior_models/base_model.py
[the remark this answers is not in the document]
annalena-k: In general, I think we should run black on all files. Although the in-between period results in messy PRs, I am hoping for cleaner PRs in the future.

annalena-k (approved):
Thank you Vincent!

dingo/core/utils/backward_compatibility.py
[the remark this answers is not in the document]
stephengreen: Agreed, we should switch to the logger in general.

stephengreen (approved):
Looks good to me! Thanks!

vincentberenz:
@stephengreen @annalena-k Cool, but it looks like I do not have the rights to perform the merge myself :)

stephengreen:
Oops I'll fix that when I get to my computer. I'll merge in the meantime.

PR #289, opened 2025-03-03
Loading polarizations from disk

annalena-k:
This PR addresses the memory issue when training a dingo model with a large waveform dataset.

Previously:
* Full waveform dataset had to be loaded into memory at the beginning of training. This required a large amount of RAM when training with a large waveform dataset and can result in problems when training multiple RAM-consuming models on one GPU node.
* Dataset was directly loaded from the path where it was stored.

Now:
* It is possible to include a temporary path in the local settings: `local_cache_path: tmp` which will result in the dataset being copied over to the local node at the beginning of training. For this purpose, the user can include `request_disk: 300GB` in the condor settings.
* The user can specify to not load the polarizations into RAM at the beginning of training, but load the required batch on-the-fly during training. To do so, one has to add this to the local settings in the `train_settings.yaml` file: `leave_waveforms_on_disk: True`

At the moment, it is only possible to load the waveforms during training, not the parameters. The reason is that we have to compute the standardization for each parameter at the beginning of training. Therefore, all parameters have to be loaded into memory. 

It would be possible to also load the parameters during training if we make these additional changes: 
(1) Computing the standardization dict before saving the waveform dataset, 
(2) Storing the standardization dict in the waveform dataset, 
(3) Loading the standardization dict at the beginning of training, and 
(4) Adapting the code to also index into the parameters.
But we leave these changes for future updates.

It is highly recommended to always copy over the waveform dataset to the local node at the beginning of training if the polarizations are left on disk. This avoids delays because of network traffic.


dingo/core/dataset.py
```diff
@@ -24,27 +25,74 @@ def recursive_hdf5_save(group, d):
...
-                d[k] = recursive_hdf5_load(v)
+                d[k] = recursive_hdf5_load(
+                    v, wfd_keys_to_leave_on_disk=wfd_keys_to_leave_on_disk, idx=idx
+                )
             else:
-                d[k] = v[...]
-                # If the array has column names, load it as a pandas DataFrame
-                if d[k].dtype.names is not None:
-                    d[k] = pd.DataFrame(d[k])
-                # Convert arrays of size 1 to scalars
```
stephengreen, line 38: The logic could be simplified quite a bit. Just insert some code here (in the old code) that takes the specific elements specified in `idx`.

dingo/gw/training/train_builders.py
```diff
@@ -29,7 +29,11 @@
 from dingo.core.utils import *
 
 
-def build_dataset(data_settings):
+def build_dataset(
+    data_settings: dict,
+    path_copy_wfd_to_local: Optional[str] = None,
```
stephengreen, line 34: It might make more sense to put the copying of the dataset in `train_pipeline_condor`, as I think it would only be used when using condor, and is logically separate from the training transforms

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -210,20 +241,98 @@ def __getitem__(self, idx) -> Dict[str, Dict[str, Union[float, np.ndarray]]]:
...
+                #   in dataset
+                # * Load standardization dict at the beginning of training
+                # * Adapt code to allow loading parameters from disk
+            elif "parameters" in self.wfd_keys_to_leave_on_disk:
+                raise ValueError(
+                    "Loading parameters from disk is not implemented at the moment because parameter"
+                    "standardization over the full dataset happens at the beginning of training."
+                    "The standardization dict could be saved when generating the waveform dataset."
+                )
+                # TODO: Adapt code to also load parameters from disk (see comment above)
```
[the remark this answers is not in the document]
annalena-k: Does this mean that you would not throw errors here and remove the different options completely?
How would you design the changes to the `train_settings.yaml` file then?

dingo/gw/training/train_builders.py
annalena-k: There are some cases where I need to copy the waveform dataset at the beginning of training, although I am using `dingo_train` instead of `dingo_train_condor`. For example, if I want to check the memory footprint of the training, I connect to an interactive node on the cluster and use `dingo_train`. Since the dataset is stored somewhere else on the cluster and I want to load the polarizations from disk (without network traffic interference), I still want to copy over the dataset to my local node.
I agree that this is not the usual use case and mainly applies to debugging GPU issues and memory errors.

However, there's another reason why I put the copy command into this function and not into `train_pipeline_condor`:
I have to overwrite the path to the waveform dataset from the permanent path to the local path, but I have to make sure that the permanent path still gets saved in the `PosteriorModel` such that resuming training works without issues.
If I would overwrite the path in `train_pipeline_condor`, this path would be stored in `PosteriorModel`. I would either need to include the additional argument `local_wfd_path` in the functions `prepare_training_new` and `prepare_training_resume` or write it into the `train_settings` dictionary. The latter option is not ideal because it would save the `local_wfd_path` in the posterior model metadata (which is completely unnecessary).

What do you think should I do?

dingo/core/dataset.py
```diff
@@ -24,27 +25,74 @@ def recursive_hdf5_save(group, d):
...
+                        sorted_data = v[sorted_idx]
+                        d[k] = sorted_data[reverse_sorting]
+                    else:
+                        # Load specific idx
+                        d[k] = v[idx]
+                    # If the array has column names, load it as a pandas DataFrame
+                    if d[k].dtype.names is not None:
+                        # Convert row v[idx] into list for pd
+                        if type(d[k]) == np.void:
+                            d[k] = pd.DataFrame([list(d[k])], columns=d[k].dtype.names)
```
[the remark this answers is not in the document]
annalena-k: This is required to format the data frame correctly if a full batch of parameters is loaded from disk.
I included it because I thought that we might load parameters from disk in the future, but since this is not the case, I removed it again.

dingo/core/dataset.py
```diff
@@ -107,11 +168,31 @@ def to_file(self, file_name, mode="w"):
...
+        else:
+            print(f"Loading dataset from {str(file_name)}.")
+        # Replace key 'polarizations' with 'h_cross' and 'h_plus'
+        if (
+            wfd_keys_to_leave_on_disk is not None
+            and "polarizations" in wfd_keys_to_leave_on_disk
+        ):
+            wfd_keys_to_leave_on_disk.remove("polarizations")
+            wfd_keys_to_leave_on_disk.append("h_cross")
+            wfd_keys_to_leave_on_disk.append("h_plus")
```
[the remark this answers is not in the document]
annalena-k: This was necessary when `wfd_keys_to_leave_on_disk` was still passed as an argument to `recursive_hdf5_load` because it iteratively goes into `polarizations` and then `h_cross`.
I removed it since this is not needed if `wfd_keys_to_leave_on_disk` is not an argument for `recursive_hdf5_load`.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -210,20 +241,98 @@ def __getitem__(self, idx) -> Dict[str, Dict[str, Union[float, np.ndarray]]]:
...
+            elif (
+                "h_cross" in self.wfd_keys_to_leave_on_disk
+                or "h_plus" in self.wfd_keys_to_leave_on_disk
```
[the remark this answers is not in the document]
annalena-k: See above

dingo/core/dataset.py
```diff
@@ -24,27 +25,74 @@ def recursive_hdf5_save(group, d):
...
+    keys: list[str] or None
+        List of keys to load. If None, load all keys.
+    wfd_keys_to_leave_on_disk: list[str] or None
+        Keys that should not be loaded into RAM when initializing the dataset. If None, all keys are loaded.
+    idx: int or list[int] or None
+        If idx is provided, only the datapoints corresponding to the given indices are loaded.
+        This functionality is needed at train time when the data corresponding to the leave_on_disk_keys
+        has to be loaded for each idx/batch.
+    """
+    non_idx_keys = ["V", "mismatches", "s"]
```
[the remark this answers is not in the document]
annalena-k: Thanks, I removed this.

dingo/core/dataset.py
annalena-k: I am not sure I understand this correctly: Do you mean that we can simplify the type conversions (data frame vs. scalar vs. string vs. list of strings)?

I don't directly see how we could simplify this because we would need an if-else to make sure that `idx` is not None since `v[idx]` with `idx=None` would fail.

dingo/gw/training/train_builders.py
annalena-k: I moved copying the waveform dataset into a separate function in `train_pipeline.py` and removed it from `train_builders.build_dataset()`.

dingo/gw/dataset/waveform_dataset.py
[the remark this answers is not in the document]
annalena-k: I replaced `wfd_keys_to_leave_on_disk` by `leave_polarizations_on_disk`.

dingo/core/dataset.py
```diff
@@ -24,27 +25,74 @@ def recursive_hdf5_save(group, d):
             raise TypeError(f"Cannot save datatype {type(v)} as hdf5 dataset.")
 
 
-def recursive_hdf5_load(group, keys=None):
+def recursive_hdf5_load(
+    group,
+    keys: Optional[List[str]] = None,
+    wfd_keys_to_leave_on_disk: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
annalena-k: I removed `wfd_keys_to_leave_on_disk` from `recursive_hdf5_load`.

dingo/core/dataset.py
annalena-k: We resolved this in a discussion.

dingo/core/dataset.py
```diff
@@ -61,7 +90,13 @@ class DingoDataset:
 
     dataset_type = "dingo_dataset"
 
-    def __init__(self, file_name=None, dictionary=None, data_keys=None):
+    def __init__(
+        self,
+        file_name: Optional[str] = None,
+        dictionary: Optional[dict] = None,
+        data_keys: Optional[List] = None,
+        leave_polarizations_on_disk: Optional[bool] = False,
```
stephengreen, line 98: We shouldn't make reference to polarizations from within `dingo.core`. How about replacing this with `leave_on_disk_keys`, which takes a list.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -51,25 +57,45 @@ def __init__(
...
+                raise TypeError(
+                    'precision can only be changed to "single" or "double".'
+                )
+        else:
+            self.real_type, self.complex_type = None, None
         super().__init__(
             file_name=file_name,
             dictionary=dictionary,
             data_keys=["parameters", "polarizations", "svd"],
+            leave_polarizations_on_disk=leave_polarizations_on_disk,
```
stephengreen, line 86: If `DingoDataset.__init__` is changed to take `leave_on_disk_keys`, then set that to `["polarizations"]` here if `leave_polarizations_on_disk=True`.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -51,25 +57,45 @@ def __init__(
...
+                self.real_type = np.float32
+            elif self.precision == "double":
+                self.complex_type = np.complex128
+                self.real_type = np.float64
+            else:
+                raise TypeError(
+                    'precision can only be changed to "single" or "double".'
+                )
+        else:
+            self.real_type, self.complex_type = None, None
```
stephengreen, line 81: This is a bit cumbersome. You could make `self.complex_type` and `self.real_type` properties of the class, e.g.,
```python
@property
def complex_type(self):
    if self.precision is not None:
        if self.precision == "single":
            return np.complex64
        elif self.precision == "double":
            return np.complex128
        else:
            raise TypeError()
    else:
        return None
```
Later your checks can be made easier. Just check whether `self.precision` is not `None`.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -90,32 +116,32 @@ def load_supplemental(self, domain_update=None, svd_size_update=None):
...
-                    'precision can only be changed to "single" or "double".'
-                )
-            self.parameters = self.parameters.astype(real_type, copy=False)
-            for k, v in self.polarizations.items():
-                self.polarizations[k] = v.astype(complex_type, copy=False)
+        if (
+            self.precision is not None
+            and self.complex_type is not None
+            and self.real_type is not None
+        ):
```
stephengreen, line 123: Only need to check that `self.precision` is not `None`, since the others are tied to it (especially if you use properties).

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -90,32 +116,32 @@ def load_supplemental(self, domain_update=None, svd_size_update=None):
...
+            if self.parameters is not None:
+                self.parameters = self.parameters.astype(self.real_type, copy=False)
+            if (
+                self.polarizations is not None
+                and self.polarizations["h_cross"] is not None
+                and self.polarizations["h_plus"] is not None
```
stephengreen, line 129: Can drop these two tests. Checking `polarizations is not None` should be enough.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -139,17 +165,20 @@ def update_domain(self, domain_update: dict = None):
...
         if (
             self.settings.get("compression", None) is not None
             and "svd" in self.settings["compression"]
         ):
             self.svd["V"] = self.domain.update_data(self.svd["V"], axis=0)
-        else:
+        elif (
+            self.polarizations["h_cross"] is not None
+            and self.polarizations["h_plus"] is not None
+        ):
```
stephengreen, line 177: I think, if you leave the polarizations on disk, then `self.polarizations = None`, so you can just check this. In fact, `self.polarizations["h_cross"]` and `self.polarizations["h_plus"]` shouldn't exist.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -210,20 +240,67 @@ def __getitem__(self, idx) -> Dict[str, Dict[str, Union[float, np.ndarray]]]:
...
+            }
+            # Convert parameters to dict
+            if not isinstance(parameters, dict):
+                parameters = parameters.to_dict()
+            # Update precision
+            if (
+                self.precision is not None
+                and self.real_type is not None
+                and self.complex_type is not None
+            ):
```
stephengreen, line 279:
```suggestion
            if self.precision is not None:
```

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -210,20 +240,67 @@ def __getitem__(self, idx) -> Dict[str, Dict[str, Union[float, np.ndarray]]]:
...
+                parameters = {
+                    k: v.astype(self.real_type, copy=False)
+                    for k, v in parameters.items()
+                }
```
stephengreen, line 283: Is this necessary? It should have happened on line 125 above.

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -210,20 +240,67 @@ def __getitem__(self, idx) -> Dict[str, Dict[str, Union[float, np.ndarray]]]:
...
+            # Perform SVD size update on waveform
+            if self.svd is not None and self.svd_size_update is not None:
+                for k, v in polarizations.items():
+                    if len(v.shape) == 1:
+                        polarizations[k] = v[: self.svd_size_update]
+                    else:
+                        polarizations[k] = v[:, : self.svd_size_update]
+        else:
+            parameters = {
+                k: v if isinstance(v, float) else v.to_numpy()
```
stephengreen, line 297:
```suggestion
                k: v.to_numpy()
```
(This makes it consistent with #290.)

dingo/gw/dataset/waveform_dataset.py
```diff
@@ -235,7 +312,7 @@ def __getitems__(
         if self.transform is not None:
             data = self.transform(data)
 
-        # Currently, the data is of shape [M, N, ...] with where M is the number
+        # Currently, the data is of shape [M, N, ...] where M is the number
```
stephengreen, line 315: Merge in the latest `main` branch to get updates to these comments from #290.

dingo/gw/training/train_pipeline.py
```diff
@@ -42,10 +74,20 @@ def prepare_training_new(train_settings: dict, train_dir: str, local_settings: d
...
+    data_settings = deepcopy(train_settings["data"])
+    if "path_copy_wfd_to_local" in local_settings:
+        data_settings["waveform_dataset_path"] = copy_files_to_local(
+            file_path=data_settings["waveform_dataset_path"],
+            local_dir=local_settings["path_copy_wfd_to_local"],
+        )
+    wfd = build_dataset(
+        data_settings=data_settings,
+        leave_polarizations_on_disk=local_settings.get(
+            "leave_polarizations_on_disk", None
```
stephengreen, line 88: This should probably be made `True` by default so the user doesn't have to think about it. I can't imagine why we wouldn't leave the waveforms on disk.

dingo/gw/training/train_pipeline.py
```diff
@@ -130,7 +176,18 @@ def prepare_training_resume(checkpoint_name, local_settings, train_dir):
...
+    data_settings = deepcopy(pm.metadata["train_settings"]["data"])
+    if "path_copy_wfd_to_local" in local_settings:
+        data_settings["waveform_dataset_path"] = copy_files_to_local(
+            file_path=data_settings["waveform_dataset_path"],
+            local_dir=local_settings["path_copy_wfd_to_local"],
+        )
+    wfd = build_dataset(
+        data_settings=data_settings,
+        leave_polarizations_on_disk=local_settings.get(
+            "leave_polarizations_on_disk", None
```
stephengreen, line 188:
```suggestion
            "leave_polarizations_on_disk", True
```

stephengreen (changes requested):
Looks good, just a couple more comments to address. Thanks for all this work!

Aside from the comments in the code, I think one could simplify a little bit some of the new keys, e.g.,
* "leave_polarizations_on_disk" -> "leave_waveforms_on_disk"
* "path_copy_wfd_to_local" -> "local_storage_path"

dingo/core/dataset.py
annalena-k: True, I corrected this.

dingo/gw/dataset/waveform_dataset.py
annalena-k: Yes, I changed this.

dingo/gw/training/train_pipeline.py
annalena-k: Yes, the only downside is that the waveform dataset should be copied to local always at the beginning of training.
I changed it, but I will include a warning when `leave_waveforms_on_disk = True` but the dataset is not copied over.

dingo/gw/training/train_pipeline.py
annalena-k: I changed this

dingo/gw/dataset/waveform_dataset.py
annalena-k: Thanks for the idea, I didn't think of this and it simplifies the code nicely! 
I included it.

dingo/gw/dataset/waveform_dataset.py
annalena-k: I included the properties and removed the additional checks.

dingo/gw/dataset/waveform_dataset.py
annalena-k: Yes, I agree, I changed it.

dingo/gw/dataset/waveform_dataset.py
annalena-k: Completely agree, changed it.

dingo/gw/dataset/waveform_dataset.py
annalena-k: Done.

dingo/gw/dataset/waveform_dataset.py
annalena-k: Done.

examples/gnpe_model/train_settings.yaml
```diff
@@ -110,9 +110,12 @@ local:
     max_time_per_run: 3600000
     max_epochs_per_run: 500
   checkpoint_epochs: 10
+  leave_waveforms_on_disk: True
+  local_cache_path: tmp
```
stephengreen, line 114: For this, and the other example files, I'm not sure we should specify `local_cache_path`, as many people would train locally, and wouldn't want to copy the dataset.

dingo/gw/training/train_pipeline.py
```diff
@@ -21,9 +27,57 @@
     build_train_and_test_loaders,
 )
 from dingo.core.utils.trainutils import EarlyStopping
+from gw.dataset import WaveformDataset
+from core.posterior_models import BasePosteriorModel
```
stephengreen, line 31:
```suggestion
from dingo.gw.dataset import WaveformDataset
from dingo.core.posterior_models import BasePosteriorModel
```
This looks like a typo. Not sure how it got here...

dingo/core/dataset.py
```diff
@@ -107,11 +146,20 @@ def to_file(self, file_name, mode="w"):
...
+    def from_file(self, file_name: str):
+        keys_to_load = self._data_keys
+        if self.leave_on_disk_keys is not None:
+            print(
+                f"Loading dataset without loading {self.leave_on_disk_keys} from {str(file_name)}."
+            )
+            # Remove keys from keys_to_load
+            keys_to_load = [k for k in keys_to_load if k not in self.leave_on_disk_keys]
+        else:
+            print(f"Loading dataset from {str(file_name)}.")
```
stephengreen, line 158: This could be slightly simplified if we make `self.leave_on_disk_keys` default to an empty list instead of None. I'll make the change, along with a couple other very minor things.

stephengreen (changes requested):
Thanks for making all these changes. Just a couple minor comments:
* The example config files should probably not copy the waveforms, since many users would run locally. If you want to include the line as an option, at least comment it out.
* It would be better to only issue the warning on not copying the waveform dataset to local if condor is being used (again since many people will run locally and want to leave waveforms on disk). This check could be moved to `train_pipeline_condor.py`.

There are also a couple minor issues / typos that I'll just address in a quick commit.

dingo/gw/training/train_pipeline.py
stephengreen: Fixed in latest commit.

dingo/core/dataset.py
stephengreen: Changed in latest commit.

examples/gnpe_model/train_settings.yaml
annalena-k: I only included `local_cache_path` for examples where `condor` is included.
You can have a look at the FMPE example: There, `condor` is not included and so is local_cache_path`.

annalena-k: If you think that most people don't train with `condor`, then we should not only comment out `local_cache_path` but also the condor settings.

dingo/gw/training/train_pipeline.py
```diff
@@ -31,25 +31,47 @@
...
+        print("Done. This took {:2.0f}:{:2.0f} min.".format(*divmod(elapsed_time, 60)))
+    elif leave_keys_on_disk:
+        print(
+            f"Warning: leave_waveforms_on_disk defaults to True, but local_cache_path is not specified. "
+            f"This means that the waveforms will be loaded during training from {local_file_path} ."
+            f"This can lead to unexpected long times for data loading during training due to network traffic. "
+            f"To prevent this, specify 'local_cache_path = tmp' in the local settings or set "
+            f"leave_waveforms_on_disk = False. However, the latter is not recommended for large datasets since "
+            f"it can lead to memory issues when loading the entire dataset into RAM. "
+        )
```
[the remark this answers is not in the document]
annalena-k: I addressed this.

examples/gnpe_model/train_settings.yaml
annalena-k: I commented it out and added the remark: 
```
# local_cache_path: tmp # uncomment to avoid slow data loading during training due to network traffic on cluster
```
such that we can finally close this PR!

annalena-k:
The last commit is a small fix for the documentation, because the `asimov` webpage is not listed in the side bar at the moment.

stephengreen (approved):
Thank you! Looks good!

PR #295, opened 2025-03-10
dingo pipe model meta data: falling back from 'meta' if not available

vincentberenz:
In a recent pull request, we added the "load_with_fallback".  Initial motivation was better support of older version of python and torch which does not support the 'meta' device.

At that time, we device to not use fallback when instantiating posterior models, as the user should be in full control of on which device models are loaded.

Issue: the dingo pip main forces instantiation of model on the 'meta' device:

```python
  model = build_model_from_kwargs(
            filename=args.model, device="meta", load_training_info=False
        )
  model_metadata = model.metadata
```

As a result the toy example still fails on older python / torch.

This pull request fallback to the device specified in the 'ini' file is meta is not supported:

```python
    try:
        model = build_model_from_kwargs(
            filename=args.model, device="meta", load_training_info=False
        )
    except RuntimeError:
        # 'meta' is not supported by older version of python / torch
        model = build_model_from_kwargs(
            filename=args.model, device=args.device, load_training_info=False
        )
  model_metadata = model.metadata
```

This change does not touch the posterior model files, so user is still in full control.

### DIFF
--- a/dingo/pipe/main.py
+++ b/dingo/pipe/main.py
@@ -6,26 +6,34 @@ import os
 from bilby_pipe.input import Input
 from bilby_pipe.main import MainInput as BilbyMainInput
 from bilby_pipe.utils import (
-    parse_args,
+    convert_string_to_dict,
     get_command_line_arguments,
     logger,
-    convert_string_to_dict,
+    parse_args,
 )
+
+from dingo.core.posterior_models.build_model import build_model_from_kwargs
+from dingo.gw.domains import build_domain_from_model_metadata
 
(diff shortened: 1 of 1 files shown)


annalena-k:
Looks good to me.

annalena-k (approved):
Looks good to me

stephengreen (approved):

vincentberenz:
@stephengreen I still do not have the rights to merge :)

annalena-k:
I merged it :) But it would be great if @stephengreen could give @vincentberenz the right to merge into main.

PR #296, opened 2025-03-11
added dingo-ci (running toy example and sending email reports)

vincentberenz:
This adds the folder "ci" (continuous integration), which contains all the files required to setup a continuous integration server which runs the toy example when a new commit is created on main; or when a new tag is added to the repo.

The files are:

- dingo-ci : a bash script which clones dingo, runs the toy example and sends a mail report
- Dockerfile: for running dingo-ci in docker
- dingo-ci-trigger: a bash script which checks the dingo repo for new commits/tags, and call dingo-ci in docker accordingly
- dingo-ci.service, dingo-ci.timer: the systemctl service which runs dingo-ci-trigger every 5 minutes
- README: more info about all this


ci/README.md
annalena-k, line 1: Could you include a few lines with instructions for what would need to be changed to also run e.g. a small version of the GNPE model?

annalena-k (commented):
Small question, otherwise looks good to me!

ci/README.md
annalena-k: Thank you, sounds reasonable to not include it!

Issue #266, opened 2024-10-25
Raise an exception when the importance sampling prior is broader than the training prior


nihargupte-ph:
When increasing the size of the prior during importance sampling, it's possible that the sampling efficiency drops to <0.1% and the evidence is very high (ln evidence = -19). This happens for the following reason: 

Suppose that one has a uniform prior in the distance from (0, 10_000) Mpc for training and then does importance sampling to (0, 15_000) Mpc. Because the flow doesn't learn exactly a hard boundary, there is some chance that samples `x` will be sampled outside of the 10_000 Mpc boundary. 

1). In [this line](https://github.com/dingo-gw/dingo/blob/main/dingo/pipe/importance_sampling.py#L159), we sample the synthetic phase using the old prior. Then the `log_prior` for `x` will be `-np.inf`. Consequently, the `log_likelihood` and `log_prob` for these will be NaN. 

2). However, in this line: [this line](https://github.com/dingo-gw/dingo/blob/main/dingo/pipe/importance_sampling.py#L185)  (in particular [here](https://github.com/dingo-gw/dingo/blob/main/dingo/gw/result.py#L238)), the prior is updated. Thus the `log_prior` for `x` is now finite. 

3). With the updated prior when [recomputing the evidence](https://github.com/dingo-gw/dingo/blob/main/dingo/core/result.py#L348), the samples of `x` have NaN `log_prob` and NaN `log_likelihood` which get cast to 0 with `np.nan_to_num`. But since the `log_prior` is now no longer `-np.inf`, the weights of `x` will orders of magnitude too high causing the appearance of low efficiency. 

The solution would be to 

A). Throw a error when the support of the new prior is non-overlapping with the support of the old prior. 

B). Evaluate `x` under the old prior. Then if `x` falls outside the old prior's support, reset the `log_prior` of `x` to `-np.inf`

The problem with A is that one can imagine a scenario where one wants to change the new prior to be a different shape than the old prior to match runs from another sampler. If the posterior has zero support over extended region of the prior, then redefining the prior is fine. I guess it then makes sense to send a warning? 

The problem with B is that I'm not entirely sure is this is correct due to the normalization. If `x` falls outside the old_prior but still is in the domain of the flow, then technically, it should not have `-np.inf` `log_prob`. After all, we were able to draw the sample. During IS if we use the old_prior this issue is fixed. The importance weight of `x` should be 0 anyway. But if we use the new_prior, then the weight shouldn't neccesarily be zero to preserve the normalization.

PR #298, opened 2025-03-12
Fixing Issue When IS prior is broader than train prior

nihargupte-ph:
The issue is described [here](https://github.com/dingo-gw/dingo/issues/266). 

This is addressed by changing the prior before synthetic phase sampling instead of afterwards. That way, the prior doesn't default to -np.inf values and the log_likelihood and log_prob are actually evaluated. 

I tested on a GW event which was giving high evidence and low sample efficiency due to the bug and after this fix I get a reasonable evidence and 7% efficiency.


stephengreen (approved):
Looks good to me. Perhaps @max-dax wants to check that he doesn't see any issues with this re-ordering.

PR #302, opened 2025-03-17
Update event_dingo.yaml

sclyne623:
Fixed typo with tabing in event_data.yaml so that "event time" "psd length" and "interferometers" are under the data settings


nihargupte-ph (approved):
LGTM

PR #303, opened 2025-03-17
Fix bug in recursive_hdf5_load

stephengreen:
When loading dataset elements batch-wise from disk, the ordering was not maintained properly. This effectively randomized the waveforms with respect to parameters.

This PR fixes the ordering.


annalena-k (approved):
I added the test that @stephengreen included in the multibanding branch in this commit: https://github.com/dingo-gw/dingo/pull/292/commits/555a2d9e3645f52580366bf157ae28d7c2af7bad
and added an additional test which compares the loaded samples for `leave_waveforms_on_disk=True` vs. `=False`.

I tested that both tests fail on main and do not fail on this branch.

PR #305, opened 2025-03-20
fix bug introduced in cvmfs integration

annalena-k:
Bug:
`dingo_pipe` fails in CI with the following error:
``` File "/data/dingo/venv/lib/python3.12/site-packages/dingo/pipe/sampling.py", line 115, in _load_sampler
    filename=self.model, device=self.device, load_training_info=False
             ^^^^^^^^^^
AttributeError: 'SamplingInput' object has no attribute 'model'
```

Reason:
- An alternative path for `self.model` is required when running on OSG. This resulted in changes [when defining self.model](https://github.com/dingo-gw/dingo/blob/24685a911ade7d46ae6a8f25295597ec3306c363/dingo/pipe/sampling.py#L49)
- However, `self.model` and `self.model_init` are not defined if `args.osg is False`, leading to the CI failure.

This PR fixes this by including an `else` option.

### DIFF
--- a/dingo/pipe/sampling.py
+++ b/dingo/pipe/sampling.py
@@ -2,7 +2,7 @@
 """ Script to sample from a Dingo model. Based on bilby_pipe data analysis script. """
 import sys
 from pathlib import Path
-import os 
+import os
 
 from bilby_pipe.input import Input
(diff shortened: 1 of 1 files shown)


nihargupte-ph (approved):
LGTM

PR #306, opened 2025-03-21
Use safer options for randomized_svd

mpuerrer:
Prevent segfaults in multi-threaded LU decomposition, by switching to the QR decomposition as the normalizer for the power iteration algorithm used in randomized_svd().

The segfault happens in when applying the LU normalizer to `A @ Q` at https://github.com/scikit-learn/scikit-learn/blob/98ed9dc73/sklearn/utils/extmath.py#L335, not in the call to the standard SVD.

### DIFF
--- a/dingo/gw/SVD.py
+++ b/dingo/gw/SVD.py
@@ -4,7 +4,6 @@ import scipy
 from sklearn.utils.extmath import randomized_svd
 from dingo.core.dataset import DingoDataset
 
-
 class SVDBasis(DingoDataset):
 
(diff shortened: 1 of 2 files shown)


stephengreen (changes requested):
Thanks! So with this change, can we remove the pinned scipy version? If so, please remove it from `pyproject.toml`.

mpuerrer:
I have removed pinning scipy now.

stephengreen (approved):
Great, thanks for tracking this down!

PR #307, opened 2025-04-01
Transfer files hotfix

nihargupte-ph:
Allowing for file transfer to condor nodes when using dingo_pipe.

### DIFF
--- a/dingo/pipe/nodes/sampling_node.py
+++ b/dingo/pipe/nodes/sampling_node.py
@@ -54,7 +54,7 @@ class SamplingNode(AnalysisNode):
 
                 # only supporting OSDF transfers for now
                 # stripping osdf prefix as it is not needed
-                network_files = [s.replace("/osdf", "") for s in [self.inputs.model, self.inputs.model_init]]
+                network_files = [s.replace("/osdf", "") for s in [self.inputs.model, self.inputs.model_init] if s is not None]
                 network_files = [f"igwn+osdf://{s}" for s in network_files]
                 input_files_to_transfer.extend(network_files)
                 # This is needed to access the networks which are in OSDF
--- a/dingo/pipe/nodes/sampling_node.py
+++ b/dingo/pipe/nodes/sampling_node.py
@@ -20,19 +20,6 @@ class SamplingNode(AnalysisNode):
 
         self.setup_arguments()
 
-        if self.inputs.transfer_files or self.inputs.osg:
-            data_dump_file = generation_node.data_dump_file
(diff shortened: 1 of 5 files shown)


dingo/pipe/nodes/sampling_node.py
```diff
@@ -53,20 +40,25 @@ def __init__(self, inputs, generation_node, dag):
...
+            ]
+            if self.inputs.osg:
+                sites = self.inputs.gpu_desired_sites
+                if sites is not None:
+                    self.extra_lines.append(f'MY.DESIRED_Sites = "{sites}"')
+                self.requirements.append("IS_GLIDEIN=?=True")
+
+                # only supporting OSDF transfers for now
+                # stripping osdf prefix as it is not needed
+                network_files = [s.replace("/osdf", "") for s in [self.inputs.model, self.inputs.model_init]]
```
annalena-k, line 57: Is this also working for NPE models, where `self.inputs.model_init = None`?

annalena-k (changes requested):
Looks good to me, in addition to the one comment, I have a question:
It is not clear to me why we need two separate ini file arguments that trigger the file transfer. In which scenario do we want to transfer files but not run on the OSG?

Maybe you can just include the explanation here: https://git.ligo.org/pe/O4/dingo-review/-/wikis/O4b-review/tutorials/OSG-tutorial

Do we also want to document this on our general dingo docs for completeness?

nihargupte-ph:
> Looks good to me, in addition to the one comment, I have a question: It is not clear to me why we need two separate ini file arguments that trigger the file transfer. In which scenario do we want to transfer files but not run on the OSG?
> 
> Maybe you can just include the explanation here: https://git.ligo.org/pe/O4/dingo-review/-/wikis/O4b-review/tutorials/OSG-tutorial
> 
> Do we also want to document this on our general dingo docs for completeness?

It happens when one is running a condor setup (not OSG) which is configured to not have access to the file system where you specify the psd dict, calibration dict, importance sampling result etc. It doesn't neccesarily have to do with the OSG. Instead of putting in in the OSG tutorial I just wrote it in `dingo.pipe.main`

dingo/pipe/nodes/sampling_node.py
nihargupte-ph: That's true, see fix in : https://github.com/dingo-gw/dingo/pull/307/commits/60d262b85fd7c21abdc1219ad5ef1019de383133

annalena-k (approved):
Looks good, thanks!

PR #310, opened 2025-04-30
leave waveforms on disk in dingo_ls for waveform datasets

annalena-k:
I ran into memory issues when trying to extract the metadata (e.g., SVD compression statistics) from a large waveform dataset with `dingo_ls`.

Explanation of the bug:
* `dingo_ls` loads the full waveform dataset into memory. This can lead to memory issues when running this command on cluster login nodes with limited memory.

Solution:
* add key `leave_waveforms_on_disk=True` to prevent waveform dataset from loading the waveforms into memory.

I tested it on our cluster and it works now.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Python 3.10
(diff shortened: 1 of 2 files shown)


stephengreen (approved):
Looks good, thanks for the fix!

PR #318, opened 2025-07-29
SVD dependency issue

nihargupte-ph:
Tests are currently failing due to the fact that `scikit-learn > 1.2` no longer supports complex datatypes when generating an SVD. This change changes the default SVD to use scipy instead. One may need to run `export OPENBLAS_NUM_THREADS=1 MKL_NUM_THREADS=1` before generating the SVD to avoid a segfault.


annalena-k (approved):
Looks good to me!

PR #322, opened 2025-09-10
Injection event dataset settings and parser fixes

alexanderroussopoulos:
This pull request is comprised of two fixes following the enabling of injections by dingo pipe. 

1. **Event file storage issue**

When the event dataset file is stored in `save_hdf5` in `dingo/pipe/data_generation.py`, in the case that the injection flag is `True` the settings dict includes values for the `optimal_SNR` and `matched_filter_SNR`. By default these are stored in this dict as `numpy.float64` and `numpy.complex128` types respectively, as opposed to `float` and `complex`, resulting in a `ValueError` when loading the event dataset file for sampling. Basic fix by adding `.item()` to the relevant variables (see change).

2. **bilby data dump argument issue**

The Boolean argument `save-bilby-data-dump`  is treated as being set to `True` regardless of the value in the config file.
This is fixed by using the `StoreBoolean` class when the argument is added to the parser in `dingo/pipe/parser.py` like other Boolean arguments.

### DIFF
--- a/dingo/pipe/data_generation.py
+++ b/dingo/pipe/data_generation.py
@@ -382,10 +382,10 @@ class DataGenerationInput(BilbyDataGenerationInput):
             # Dingo and Bilby have different geocent_time conventions.
             settings["injection_parameters"]["geocent_time"] -= self.trigger_time
             settings["optimal_SNR"] = {
-                k: v["optimal_SNR"].item() for k, v in self.interferometers.meta_data.items()
+                k: v["optimal_SNR"].item() if hasattr(v['optimal_SNR'], 'item') else v['optimal_SNR'] for k, v in self.interferometers.meta_data.items()
             }
             settings["matched_filter_SNR"] = {
-                k: v["matched_filter_SNR"].item()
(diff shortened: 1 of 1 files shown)


nihargupte-ph:
Probably yes it's only affecting certain versions of bilby_pipe. When I tested the PR it was with `bilby_pipe==1.5.0` (I think though maybe my environment has changed since then so I'm not super sure, would need to spend more time to track down). 

I guess one could check if the SNR attributes are scalars or numpy objects then depending on that call .item() or not. That way it would work with both versions of bilby_pipe. E.g. something like this : 

```
settings["optimal_SNR"] = {
    k: (v["optimal_SNR"].item() if hasattr(v["optimal_SNR"], "item") else v["optimal_SNR"])
    for k, v in self.interferometers.meta_data.items()
}
settings["matched_filter_SNR"] = {
    k: (v["matched_filter_SNR"].item() if hasattr(v["matched_filter_SNR"], "item") else v["matched_filter_SNR"])
    for k, v in self.interferometers.meta_data.items()
}
```

nihargupte-ph (approved):
It looks good to me!

PR #323, opened 2025-09-30
Update pipe for compatibility with bilby_pipe == 1.7.0.

stephengreen:
Updates dingo-pipe for compatibility with latest bilby_pipe (1.7.0).

### Changes

I looked at the bilby_pipe diff and copied various changes. In cases where we do not use the updated code, I commented it. It was necessary to made some small additional changes to ensure that the code runs (e.g., making `self.detectors` available in for ImportanceSamplingInput).

I also removed the pin to `bilby_pipe == 1.4.2`.

### Testing

I tested one analysis of an event and performed one injection. @nihargupte-ph You may wish to perform additional tests.

### To-do

@nihargupte-ph I saw that the lines you added to `MergeNode` in #307 now exist in a similar form in `bilby_pipe`, https://git.ligo.org/lscsoft/bilby_pipe/-/blob/master/bilby_pipe/job_creation/nodes/merge_node.py#L17-34. 

With the latest `bilby_pipe`, the code now executes **both** bits of code, since our `MergeNode.__init__()` calls its `super().__init__()`.  What is the appropriate behavior?

* If you want to use only our code, then we should not call the superclass init function, but then we need to copy the rest of the init function into ours.
* If you're happy to use only their code, then remove the lines from ours.
* If you're happy for both to run leave as is.
* If it's possible to use theirs by modifying some variables, then remove the lines from ours but modify the variables appropriately.

Happy to advise if you explain what this does.

### DIFF
--- a/dingo/pipe/data_generation.py
+++ b/dingo/pipe/data_generation.py
@@ -47,6 +47,9 @@ class DataGenerationInput(BilbyDataGenerationInput):
         self.ini = args.ini
         self.transfer_files = args.transfer_files
 
+        # Global settings
+        # self.cosmology = args.cosmology
+
         # Run index arguments
         self.idx = args.idx
         self.generation_seed = args.generation_seed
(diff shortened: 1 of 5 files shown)


nihargupte-ph (approved):
I think this looks good to me. If you were able to run injections and real events I don't think there is anything else for me to test.

stephengreen:
Well, we have never attempted to change the detectors in IS, and I'm not sure we want to try, so it's fine to raise an error. On the other hand, as it stands now the logger will tell the user about this update, so I think it's fine.

PR #332, opened 2025-11-03
Adding support for SEOBNRv5EHM

nihargupte-ph:
This PR adds support for SEOBNRv5EHM through the new waveform interface. Note that unlike SEOBNRv5PHM and SEOBNRv5HM, SEOBNRv5EHM uses different conditioning routines. In particular, it doesn't use the stepping back of the starting frequency since the reference frequency is equal to the starting freqeuncy. This is why different routines are use when doing the summation in the m modes.


lorenzopompili00:
Looks good! I left a few small comments. 

On a longer term, I would love for this code not to have any line like
```
if self.approximant_str == NAME
```
but to look up relevant information from the gwsignal metadata, using [`gwsignal_get_waveform_generator`](https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/python/lalsimulation/gwsignal/models/__init__.py#L7), so that the code works with any eccentric model, and one does not need to keep updating it when a new one is added. One can (in principle) check if the waveform is, e.g., eccentric, with
```
wf_gen = gwsignal_get_waveform_generator(self.approximant_str)
if wf_gen.metadata["type"] == "eccentric":
```
and `wf_gen.metadata["f_ref_spin"]`, `wf_gen.metadata["f_ref_ecc"]` specify if spins and/or eccentricity can be defined at an arbitrary `f_ref`, which could be used to automatically determine which conditioning routine is appropriate.

Still, I realize things are not too polished on the gwsignal side so it's probably not worth thinking now about these kind of changes ([TEOBResumS](https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/2496) has a few more info, but SEOBNRv5EHM is still lacking some of these); I'm leaving this comment anyway to keep this in mind, one can update this in the future.

PR #333, opened 2025-11-05
ENH: Add triangular interferometer

filippo-santoliquido:
Hi,

I am opening this pull request to enable Dingo to work automatically with a triangular interferometer configuration, such as the Einstein Telescope (ET).

I am aware that the current implementation is not particularly elegant. In its present form, creating a [new interferometer object in Bilby](https://github.com/bilby-dev/bilby/tree/main/bilby/gw/detector/detectors) also requires modifying the sections of code shown below, which is not ideal. The main issue is that Dingo currently expects `data_settings["detectors"]` to contain the full list of interferometers, and this list is directly passed to `InterferometerList`. However, doing so results in an error from Bilby, since the interferometer `'ET-1'` does not exist.  

This happens because when `InterferometerList` is called with a triangular configuration (i.e., just `'ET'`), it automatically generates all the individual interferometers defined in `data_settings["detectors"]`.

I would appreciate any suggestions for a cleaner or more flexible solution.

Thanks a lot in advance, 
Kind regards,  
Filippo

### DIFF
--- a/dingo/gw/training/train_builders.py
+++ b/dingo/gw/training/train_builders.py
@@ -104,7 +104,12 @@ def set_train_transforms(wfd, data_settings, asd_dataset_path, omit_transforms=N
 
     ref_time = data_settings["ref_time"]
     # Build detector objects
-    ifo_list = InterferometerList(data_settings["detectors"])
+    if len(data_settings["detectors"]) == 3 and all(d.startswith('ET') for d in data_settings["detectors"]):
+        ifo_list = InterferometerList(['ET'])
+    elif all(d.startswith('ET') for d in data_settings["detectors"][:3]) and len(data_settings["detectors"]) > 3:
+        ifo_list = InterferometerList(['ET'] + data_settings["detectors"][3:])
+    else:
+        ifo_list = InterferometerList(data_settings["detectors"])
 
     # Build transforms.
     transforms = [
--- a/dingo/gw/training/train_builders.py
+++ b/dingo/gw/training/train_builders.py
@@ -86,12 +86,16 @@ def set_train_transforms(wfd, data_settings, asd_dataset_path, omit_transforms=N
     if omit_transforms is not None:
         print("Omitting \n\t" + "\n\t".join([t.__name__ for t in omit_transforms]))
 
+    # Build detector objects
+    ifo_list = InterferometerList(data_settings["detectors"])
+    ifo_names = [ifo.name for ifo in ifo_list]
+    
     # By passing the wfd domain when instantiating the noise dataset, this ensures the
     # domains will match. In particular, it truncates the ASD dataset beyond the new
     # f_max, and sets it to 1 below f_min.
     asd_dataset = ASDDataset(
         asd_dataset_path,
-        ifos=data_settings["detectors"],
+        ifos=ifo_names,
         precision="single",
         domain_update=wfd.domain.domain_dict,
     )
@@ -103,13 +107,6 @@ def set_train_transforms(wfd, data_settings, asd_dataset_path, omit_transforms=N
         data_settings["inference_parameters"] = default_inference_parameters
 
     ref_time = data_settings["ref_time"]
-    # Build detector objects
-    if len(data_settings["detectors"]) == 3 and all(d.startswith('ET') for d in data_settings["detectors"]):
-        ifo_list = InterferometerList(['ET'])
-    elif all(d.startswith('ET') for d in data_settings["detectors"][:3]) and len(data_settings["detectors"]) > 3:
-        ifo_list = InterferometerList(['ET'] + data_settings["detectors"][3:])
-    else:
-        ifo_list = InterferometerList(data_settings["detectors"])
 
     # Build transforms.
     transforms = [
@@ -175,7 +172,7 @@ def set_train_transforms(wfd, data_settings, asd_dataset_path, omit_transforms=N
         )
     )
     transforms.append(
-        RepackageStrainsAndASDS(data_settings["detectors"], first_index=domain.min_idx)
+        RepackageStrainsAndASDS(ifo_names, first_index=domain.min_idx)
     )
     if "random_strain_cropping" in data_settings:
         transforms.append(
@@ -336,7 +333,7 @@ def build_svd_for_embedding_network(
     print(f"Truncating SVD matrices below index {wfd.domain.min_idx}.")
     print("...V matrix shapes:")
     V_rb_list = []
-    for ifo in data_settings["detectors"]:
+    for ifo in ifo_names:
         V = basis_dict[ifo].V
         assert np.allclose(V[: wfd.domain.min_idx], 0)
         V = V[wfd.domain.min_idx :]
--- a/dingo/gw/training/train_builders.py
+++ b/dingo/gw/training/train_builders.py
@@ -333,7 +333,7 @@ def build_svd_for_embedding_network(
     print(f"Truncating SVD matrices below index {wfd.domain.min_idx}.")
     print("...V matrix shapes:")
     V_rb_list = []
-    for ifo in ifo_names:
+    for ifo in ifos:
         V = basis_dict[ifo].V
         assert np.allclose(V[: wfd.domain.min_idx], 0)
         V = V[wfd.domain.min_idx :]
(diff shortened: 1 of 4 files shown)


filippo-santoliquido:
Hi,

This PR addresses a limitation in `dingo` when working with the triangular Einstein Telescope (ET).

**Background**

Currently, `dingo` assumes that the detector names in `data_settings["detectors"]` match exactly the detectors produced by `InterferometerList(data_settings["detectors"])`. This holds for standard interferometers, but breaks for a triangular ET: when `[ET]` is passed to `bilby`'s `InterferometerList`, it automatically expands into three separate detectors.

**Proposed fix**

The cleanest solution is to decouple the two concepts:

- Pass `data_settings["detectors"]` (or `metadata["train_settings"]["data"]["detectors"]`) only to `InterferometerList` for initialization.
- Retrieve the actual detector names used throughout the code dynamically via `ifo_names = [ifo.name for ifo in self.ifo_list]`, as already done in [`[injection.py#L310](https://github.com/dingo-gw/dingo/blob/45dbe10982eddc177aa3f5eaf3bf5a4d87145df3/dingo/gw/injection.py#L310)`](https://github.com/dingo-gw/dingo/blob/45dbe10982eddc177aa3f5eaf3bf5a4d87145df3/dingo/gw/injection.py#L310).

This PR updates `train_builders.py` accordingly, building on my previous PR.

**Next steps**

If this approach looks good, we should audit all places in the codebase where `settings["detectors"]` (or similar) is accessed directly under the assumption that it reflects the full list of used detectors — and replace those with `ifo_names` where appropriate.

Happy to discuss further.

Best,
Filippo

nihargupte-ph:
Hi @filippo-santoliquido, I think this is a good change and makes sense. Currently, I am not approving due to the "Next steps" which I agree with. Is there anywhere else in the code that this should be updated?

filippo-santoliquido:
Hi @nihargupte-ph, thanks a lot for your reply. I searched for `settings["detectors"]` across the entire code and found a few additional places where it seems to be used incorrectly. In particular, here: https://github.com/dingo-gw/dingo/blob/03da18a38748e4887e74c09ae3a404984cb2b40d/compatibility/update_saved_model_new_gnpe.py#L32, the variable `ifo` should instead iterate over `ifo_names`. I also think the usage of `["train_settings"]["data"]["detectors"]` should be carefully reviewed and fixed wherever necessary.

nihargupte-ph:
> Hi @nihargupte-ph, thanks a lot for your reply. I searched for `settings["detectors"]` across the entire code and found a few additional places where it seems to be used incorrectly. In particular, here:
> 
> https://github.com/dingo-gw/dingo/blob/03da18a38748e4887e74c09ae3a404984cb2b40d/compatibility/update_saved_model_new_gnpe.py#L32
> 
> , the variable `ifo` should instead iterate over `ifo_names`. I also think the usage of `["train_settings"]["data"]["detectors"]` should be carefully reviewed and fixed wherever necessary.

Sorry for the delay I missed this in my inbox.. Gotcha, I agree with your statement. I think in any case where ifo_names is called one should not iterate over the `"train_settings"]["data"]["detectors"]` but rather over the `ifo_names` via `[ifo.name for ifo in InterferometerList]`

dingo/gw/training/train_builders.py
```diff
@@ -331,7 +333,7 @@ def build_svd_for_embedding_network(
     print(f"Truncating SVD matrices below index {wfd.domain.min_idx}.")
     print("...V matrix shapes:")
     V_rb_list = []
-    for ifo in data_settings["detectors"]:
+    for ifo in ifo_names:
```
lorenzopompili00: Here `ifo_names` should be `ifos`

lorenzopompili00:
The line here: https://github.com/filippo-santoliquido/dingo-ET/blob/feature/add-triangular-interferometer/dingo/gw/inference/gw_samplers.py#L319

should be
```
ifos=[ifo.name for ifo in InterferometerList(self.detectors)],
```
(e.g., (`ET1, ET2, ET3` rather than 'ET').

lorenzopompili00:
This line here is also currently failing: https://github.com/filippo-santoliquido/dingo-ET/blob/feature/add-triangular-interferometer/dingo/pipe/main.py#L179

The problem is a naming mismatch: bilby expands `ET` -> `ET1/ET2/ET3`, so the ASD dataset is keyed by the three arms, but bilby treats `ET` as a single detector. So the PSD loop in `fill_in_arguments_from_model` does `asd_dataset.save_psd(..., 'ET')` which gives `KeyError: 'ET'`. We can't just key psd_dict by `ET1/ET2/ET3` either, because bilby_pipe's `_validate_psd_dict` checks against valid Bilby detectors, and only `ET` is a valid option.

A potential fix is to read one arm's ASD as the representative but keep the psd_dict key as 'ET', e.g.:
```python
for ifo_name in args.detectors:
    if ifo_name in asd_dataset.asds:
        asd_key = ifo_name            # normal detector, e.g. 'H1'
    else:
        asd_key = [ifo.name for ifo in InterferometerList([ifo_name])][0]  # 'ET'-> 'ET1'
    psd_path = asd_dataset.save_psd(args.outdir, asd_key, rng=rng)
    psd_dict[ifo_name] = str(psd_path)  # keyed by 'ET'
```

One thing to confirm: this saves a single PSD, assuming it is the same across the three arms. That is consistent with how ET is currently defined in bilby (all three arms use the same PSD), so I think it's probably fine, do you think this makes sense @filippo-santoliquido? If we want independent per-arm PSDs it won't be enough however (though that likely needs a fix in bilby/bilby_pipe as well to understand ET1/ET2/ET3 as separate detectors).

lorenzopompili00:
Thank you @filippo-santoliquido!

I have tried running a quick end-to-end test for an injection based on https://github.com/dingo-gw/dingo/pull/386 (changing `detectors = [H1, L1]` to `detectors = [ET]`) and with the fixes I commented above it managed to run successfully.

filippo-santoliquido:
Hi @lorenzopompili00! Thanks a lot for your work! I believe that in the future we'll need to implement ET with different PSDs for each detector, to make our simulations as close to reality as possible. In any case, we'll need to fix bilby first, as you said, and this fix will then be automatically inherited by dingo.

filippo-santoliquido:
Dear @lorenzopompili00 ! Thanks a lot again for your thorough review. I should have implemented all your comments.

lorenzopompili00 (changes requested):

dingo/pipe/main.py
```diff
@@ -176,6 +176,10 @@ def fill_in_arguments_from_model(args, perform_arg_checks=True):
         psd_dict = {}
         rng = np.random.default_rng(args.generation_seed)
         for ifo_name in args.detectors:
+            if ifo_name in asd_dataset.asds:
+                asd_key = ifo_name # normal detector, e.g. 'H1'
+            else:
+                asd_key = [ifo.name for ifo in InterferometerList([ifo_name])][0]  # 'ET'-> 'ET1' and fixing 'ET1'
             psd_path = asd_dataset.save_psd(args.outdir, ifo_name, rng=rng)
```
lorenzopompili00: Thank you @filippo-santoliquido! Just one last tiny change: here ifo_name -> asd_key (e.g., in the ET case, this should be ET1 rather than ET, otherwise it fails because of the missing key)

lorenzopompili00 (changes requested):

lorenzopompili00 (approved):

lorenzopompili00:
@filippo-santoliquido I mentioned this PR briefly at the call earlier, and it was suggested to a some comment about the lack of independent per-arm PSDs, just for future reference. For example [here](https://github.com/dingo-gw/dingo/pull/333/changes#diff-515284483d384826fc048221418b6a13f35267ecced97f6193aaa43351668524R182), something like

```
# Triangular detector ('ET'): the ASD dataset is keyed by arm name
# (ET1/ET2/ET3), but bilby_pipe only accepts 'ET' as a psd_dict key,
# so we save the first arm's PSD and use it for all three arms.
# Note: different PSDs per-arm are NOT supported. This matches 
# bilby's current ET definition (all arms share one sensitivity curve); 
# supporting this would require bilby/bilby_pipe to treat ET1/ET2/ET3 
# as separate detectors.
```
I think that after this we can merge this!

lorenzopompili00 (approved):

PR #334, opened 2025-11-05
ENH: Add signal in injections

filippo-santoliquido:
Hi,

This pull request makes the injected signal saved by default, which can be useful, for instance, when evaluating the optimal SNR.

Kind regards,
Filippo

### DIFF
--- a/dingo/gw/injection.py
+++ b/dingo/gw/injection.py
@@ -396,6 +396,7 @@ class Injection(GWSignal):
             self.whiten = False
 
         data = {}
+        h = {}
         for ifo, s in signal["waveform"].items():
             noise = (
                 (np.random.randn(len(s)) + 1j * np.random.randn(len(s)))
@@ -406,6 +407,7 @@ class Injection(GWSignal):
             data[ifo] = self.data_domain.update_data(d, low_value=0.0)
 
         signal["waveform"] = data
+        signal["h"] = h
         return signal
 
     def random_injection(self):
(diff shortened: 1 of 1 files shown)


cmfabbri (changes requested):
Hi Filippo, thank you!

I noticed that you also pushed some changes to `result.py`, adding the functions `get_one_dimensional_median_and_error_bar` and `_percentile`. Was that intentional? If not, can you please remove them from this PR?

filippo-santoliquido:
Hi @cmfabbri! That was intentional, not an oversight. I think the features added to `result.py` are genuinely useful for the code — though you're right that including them into this same PR might not be the best approach. If you're on board with the changes themselves, I don't think this PR needs further modification. Let me know your thoughts.

cmfabbri:
Can you please split the `result.py` out into a separate PR then?

As for the addition of the signal without noise, it looks fine to me. Have you tried to do an end-to-end run to see if it breaks the code anywhere?

filippo-santoliquido:
Hi @cmfabbri! I've trimmed this PR down to just the signal addition and opened [a separate one](https://github.com/dingo-gw/dingo/pull/396) for the `result.py` change (the one-dimensional median and error bar).

To answer your question: I don't think saving the signal breaks anything — it's just extra information stored in the context. I also used this feature in my own fork of `dingo` while working on [this paper](https://arxiv.org/pdf/2504.21087) and ran into no issues with it.

cmfabbri (approved):

PR #335, opened 2025-11-17
Dont transfer entire models for data generation

AlexandreGoettel:
This formats the metadata saved inside of the model as a string and saves it in the .ini file. This way, the model metadata can be accessed basically automatically.

Addresses #327


AlexandreGoettel:
I tested this locally and as a condor job on CIT, but I'm not sure how to report results.

AlexandreGoettel:
ping @stephengreen 
If this looks good for the data generation step, I'm happy to copy the same logic to other transfers.

dingo/pipe/nodes/generation_node.py
```diff
@@ -56,19 +56,6 @@ def __init__(self, inputs, trigger_time, idx, dag, parent=None, importance_sampl
                 if (value := getattr(self.inputs, attr)) is not None:
                     input_files_to_transfer.append(os.path.abspath(str(value)))
 
-            # modifying the paths for OSDF networks
```
nihargupte-ph, line 59: Why are we removing this?

nihargupte-ph (changes requested):
Thanks for putting in the PR! 12 eyes looked at this PR. 

Just one comment on the PR.

dingo/pipe/nodes/generation_node.py
AlexandreGoettel: The generation node only needed the network files to be transferred so it could access their metadata. Since it can now get the metadata from the ini file, this removes the networks from the list of files to be transferred

ewencedr:
CI is failing

PR #346, opened 2026-01-15
Fix double memory usage during dataset loading with precision conversion

### LINKED ISSUE #243: Memory usage at start of training
`dingo_train` and `dingo_train_condor` seem to use roughly double the necessary amount of CPU RAM at the start of training. This means that we must request much more memory than is ultimately necessary for training.

I suspect this occurs when we call `split_dataset_into_train_and_test()`:

https://github.com/dingo-gw/dingo/blob/bf2c41f7630860268f4e4aed8c0a6fe9d30a13a4/dingo/core/utils/torchutils.py#L232

stephengreen:
## Summary
- Add `dtype_map` parameter to `recursive_hdf5_load()` enabling direct dtype conversion during HDF5 read via h5py's astype view
- This avoids creating intermediate copies when loading datasets with precision changes (e.g., complex128 -> complex64), which was causing double memory usage
- Supports nested dtype maps for groups with mixed types (e.g., SVD has complex V and real s)

Fixes #243

## Test plan
- [x] All existing tests pass (84 tests)
- [x] Added new tests for dtype conversion functionality:
  - `test_precision_dtype_conversion`: verifies single/double precision dtypes for parameters, polarizations, and SVD
  - `test_no_precision_preserves_original_dtype`: verifies original dtypes when precision is not specified
  - `test_precision_conversion_values_unchanged`: verifies values are preserved within precision tolerance
- [x] Manual verification with large dataset showing ~24 GB peak memory vs ~96 GB previously

### DIFF
--- a/dingo/core/dataset.py
+++ b/dingo/core/dataset.py
@@ -29,6 +29,8 @@ def recursive_hdf5_load(
     group,
     keys: Optional[List[str]] = None,
     idx: Optional[Union[int, List[int]]] = None,
+    dtype_map: Optional[dict] = None,
+    _inherited_dtype=None,
 ):
     """This is a generic helper function to recursively load data from an HDF5 file.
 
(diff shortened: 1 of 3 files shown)


max-dax (approved):

dingo/core/dataset.py
```diff
@@ -29,6 +29,8 @@ def recursive_hdf5_load(
     group,
     keys: Optional[List[str]] = None,
     idx: Optional[Union[int, List[int]]] = None,
+    dtype_map: Optional[dict] = None,
+    _inherited_dtype=None,
```
vincentberenz: typing hints for _inherited_dtype missing.

vincentberenz (changes requested):
Some minor comments regarding typing hints

nihargupte-ph (approved):

vincentberenz (approved):

PR #358, opened 2026-02-26
Multi-GPU training (DDP)

annalena-k:
### Multi-GPU training with PyTorch DDP
                                                                                                                                                                        
**Summary**                                                   
This PR adds data-parallel multi-GPU training to Dingo using https://pytorch.org/docs/stable/notes/ddp.html. It also adds gradient accumulation and automatic mixed precision (AMP) as complementary features for memory-efficient training. Together these allow significantly faster training on multi-GPU nodes and greater flexibility in effective batch size.

### What's new
**Multi-GPU training (DDP)**
  - Setting `num_gpus > 1` in `local`: spawns one process per GPU via `torch.multiprocessing`. Each GPU processes a non-overlapping shard of the mini-batch; gradients are averaged across GPUs via NCCL before each parameter update.
  - `BatchNorm` layers are automatically converted to `SyncBatchNorm` so statistics are computed correctly across all GPUs.
  - The `DistributedSampler` ensures each GPU sees a different, non-overlapping slice of data each epoch.
  - Checkpoints are written only by rank 0 and are always saved without the DDP module prefix, so they can be loaded on any number of GPUs (or a single GPU / CPU) without modification.
  - Loss and timing statistics are aggregated across ranks so the training log is identical to the single-GPU case.

**Gradient accumulation**
  - New per-stage setting `gradient_updates_per_optimizer_step` (default: 1). Setting this to $k$ accumulates gradients over $k$ forward-backward passes before calling the optimizer, simulating an effective batch size of $k \times$ `batch_size` without extra GPU memory.

**Automatic mixed precision (AMP)**
  - New per-stage setting `automatic_mixed_precision` (default: False). When enabled, the forward pass runs in FP16 while optimizer state and parameter updates stay in FP32. This roughly halves GPU memory usage and increases throughput on GPUs with Tensor Core hardware (A100, V100, H100).

`dingo_train_condor`
  - `request_gpus` in the HTCondor submission file is now set automatically from `local.num_gpus`. Users no longer need to specify it in the `condor` block.

###  Settings file changes

New optional keys in local:
 ```
 local:
    device: cuda
    num_gpus: 4        # NEW — default 1 (single GPU). Set >1 for DDP.
    num_workers: 8
```
New optional keys per training stage
  ```
training:
    stage_0:
      batch_size: 512           # Always the TOTAL effective batch size across all GPUs.
      gradient_updates_per_optimizer_step: 2   # NEW — default 1
      automatic_mixed_precision: True          # NEW — default False

```
`condor` block: `num_gpus` removed
`num_gpus` is no longer needed (or expected) inside `condor`. It is injected automatically from `local.num_gpus` when building the submission file:
```
# Before (old):
  local:
    condor:
      num_gpus: 4   # ← remove this

 # After (new):
  local:
    num_gpus: 4     # ← specify here instead
    condor:
      ...           # num_gpus no longer needed here

```

### Backward compatibility
  - `num_gpus` under `local.condor`: Existing settings files that specify `num_gpus` inside the `condor` block continue to work. A DeprecationWarning is raised at runtime prompting users to move the key to `local.num_gpus`. This fallback will be removed in a future release.
  - Single-GPU / CPU: No changes to existing behaviour. The `num_gpus` key defaults to 1 if absent.
  - Checkpoints: All existing checkpoints load without modification. The new iteration counter (total optimizer steps) is read with a default of 0 for old checkpoints that lack it.
  - `batch_size` semantics: batch_size has always been documented as the total effective batch size. This PR enforces that interpretation — when `num_gpus > 1`, each GPU receives `batch_size / num_gpus samples`. No change for single-GPU users.

### Requirements
  - PyTorch ≥ 2.0 (required for `torch.amp.GradScaler` and `torch.amp.autocast`; the deprecated `torch.cuda.amp` namespace is no longer used)
  - NCCL-capable PyTorch build for multi-GPU (standard for any CUDA-enabled install)

### Tests
  - `tests/core/test_multi_gpu.py` — covers `get_num_gpus` (including deprecation warning), `build_train_and_test_loaders` with and without DDP samplers, `LossInfo` and `RuntimeLimits` in a real two-process `gloo` group, DDP state-dict stripping, and utility functions (`set_seed_based_on_rank`, `replace_BatchNorm_with_SyncBatchNorm`).
  - `tests/core/test_amp.py` — covers AMP imports, `train_epoch` without AMP (including gradient accumulation), and `train_epoch` with AMP on CUDA (skipped if no GPU).
  
Integration tests on an htcondor cluster with the NPE example:
- `dingo_train` on 1 GPU: Training works on A100 8 GPU node with `num_gpus: 1`
- `dingo_train` on 8 GPUs: Training works
- `dingo_train_condor` on 1 GPU: Training works
- `dingo_train_condor` on 8 GPUs: Training works

To test the performance, I investigated how the training time per epoch scales with the batch size on an NVIDIA A100 8-GPU node with 80 GB GPU memory:
multi_gpu_time_per_epoch
multi_gpu_time_per_epoch_log
- With a single GPU, we cannot increase the batch size beyond 16384 due to limited GPU memory.
- With 8 GPUs, we can increase the total effective batch size beyond 16384 since each GPU only processes `batch_size / num_gpus`. However, the overhead from gradient accumulation leads to a scaling which is worse than the ideal projection given by 
$\mathrm{time}(1\mathrm{GPU}) \cdot \frac{\mathrm{batchsize}(1\mathrm{GPU})}{\mathrm{batchsize}(8\mathrm{GPUs}})$
- Automatic mixed precision decreases the run time slightly on multiple GPUs.


docs/source/training_multi_gpu.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+Each value is the current step; the value in parentheses is the running average
+over the epoch. **In DDP mode all three timings are reported as the maximum across
+all GPUs** (`dist.ReduceOp.MAX`), i.e. they reflect the slowest rank.
+
+### Time Dataloader
+Wall-clock time spent loading and pre-processing one batch (waveform
+decompression, projection to detectors, noise generation). Expect this to be
+smaller than in single-GPU mode because each GPU loads a smaller per-GPU batch.
```
nihargupte-ph, line 187: This is just out of curiousity, but why is the per GPU batch smaller? Shouldn't the amount of samples on each GPU remain the same if you increase the batch size proportionally to the number of GPUs?

nihargupte-ph:
I'm reviewing this PR at the moment and will leave a more detailed response on the code soon. But for now, I was able to reproduce your plots but with 4 GPUs. Strangely, the time per epoch on my machine also increased at the last test point!

image

Interestingly, if I replace the Syncbatchnorms with just Batchnorms then the time for epoch does go down a bit more. I think this is because the Syncbatchnorm is not asynchronous unlike the gradient accumulation. This is what I'm showing in the light blue curve above and in red is the current PR. I think the grad accumulation will begin summing the gradients at the beginning of the graph while the gradients for the end of the graph are still being computed, but Syncbatchnorm has to wait. 

One possibility we discussed which could be causing the suboptimal scaling was if increasing the batch size wasn't helping because we were compute bound. However, in my experiments I tracked the `FLOPs/possible FLOPs` and I found this hovered between 60-80%: 

image

so I don't think we're compute bound yet. I am now investigating why we're not getting 100% utilization.

stephengreen:
Thanks. What if you replace BatchNorm with LayerNorm everywhere? No syncing.

annalena-k:
Thanks Nihar!
I only included the SynchBatchNorm to ensure that nothing goes wrong for users, but I never used it myself.
If we remove this, I would just add an error message to let users know that BatchNorm doesn't work with multiple GPUs.
Happy to change this if people think this is the way forward :)

nihargupte-ph:
Right, I think batch norm or layer norm could both work. When computing the syncbatchnorm you sync the batch statistics across all the GPUs. This results in two slowdowns: 

1). The communication between each GPU to compute the batch statistics
2). The fact that the GPUs can't run fully async because they need to wait for each other to reach the syncbatch layer. 

Of course this is no different than an all reduce. However, the all reduce latency is hidden because the backward pass is occurring asynchronously with the all reduce. However, in the case of the syncbatchnorm, you can't finish the forward pass until you have batch statistics and therefore there is no async. 

In our case, the batches are large enough that I think we can get away with just BatchNorm. Layernorm could also work. The only thing is we haven't tested our enets using LayerNorm. What do people think, should we use LayerNorm going forward? This seems like a slightly larger change than just switching to BatchNorm.

nihargupte-ph:
I was able to find a way to speed up multi and single-GPU training further while profiling the multi-GPU. In particular, I noticed the GPU utilization was not reaching 100% even when the batch size is maxxing out the high bandwidth memory out. This implies we are not compute bound. This can be traced to the following reasoning

1). The neural spline flow fires 25k CUDA kernels per forward pass. 
2). Each kernel *launch* costs 5 micro-s
3). Each kernel *runs* for 10 micro-s
4). Therefore, T_kernel_launch (5 micro-s * 25k) < T_kernel_run (A100's capacity)

Note this is not related to the dataloader. This is after the data is on the HBM, how much are we utilizing the GPUs FLOPs. So the GPU is finishing computations faster than they can be launched! The solution is kernel fusing using `torch.compile(pm.network)`. This reduces the number of kernels used to 10k. The GPU may still be finishing faster (which is maybe why I'm not getting to 100% utilization on my experiments), but it is faster. One could also up the intensity of the computation (make the kernels take longer). 

The problem is `torch.compile` requires the entire computation graph to have static array shapes. This requires a change to glasflows. We could also consider switching to zuko. 

The speedup with `torch.compile` is not negligible. It can be 2x. The speedup after `torch.compile` is the green. In red is after using batchnorm instead of syncbatchnorm

```
  barrier()                    # all 4 ranks aligned
  cuda.synchronize(); t0 = perf_counter()
    forward, backward         # backward includes the DDP gradient all-reduce 
    optimizer.step()
  cuda.synchronize(); dt = perf_counter() - t0
```

Then: 
```
  step_time  = median(dt over steps 20–120)          # 20 warmup steps discarded
  epoch_time = step_time * num_batches # num batches is 10 million / effective batch size
```

pr358_scaling

pr358_util

pr358_memory

A general point is that I guess we haven't really gone into the weeds of profiling how much we're getting out of the GPUs if we're maxing out the FLOPS.

nihargupte-ph:
One final point is that it would be interesting to do a study to determine what is the optimal batch size for our problem. Of course we could use as many GPUs as possible but at a certain batch size, your convergence speed starts to drop off. This was explored in https://arxiv.org/abs/1812.06162. When we start launching large training runs, this would help us determine how many GPUs we should really be using.

nihargupte-ph:
Regarding the above three points, I am fine to either 

A). submit my review for the current PR and then we merge into main. Then we can consider these optimizations later.

B). Take into account either the point about removing the syncbatchnorm or using torch.compile and directly merging into this PR. I think the third point about the batch sizes will require more time and is outside of the scope of this PR. 

Just let me know what your preference/thoughts on the above optimizations are @annalena-k @stephengreen . I think either could work, but I'm leaning more towards B to keep things organized.

dingo/core/utils/trainutils.py
```diff
@@ -96,57 +99,147 @@ def __call__(self, val_loss: float):
...
+        self.iteration += 1
+        loss = torch.mean(torch.tensor(self.cached_losses, device=self.device))
+        n = torch.tensor(sum(self.cached_n), device=self.device, dtype=torch.float32)
+
+        if self.is_ddp:
+            # Reduce absolute loss across GPUs so that normalization is correct
+            # even when GPUs process different numbers of samples.
+            abs_loss = loss * n
+            dist.barrier()
+            dist.reduce(abs_loss, dst=0)
```
nihargupte-ph: Shouldn't this be an allreduce? Right now only GPU 0 and GPU 1 would have different losses. I think this could be important if the LR changes based off the loss because then different gpus would have different losses leading to different LR schedules.

dingo/core/posterior_models/base_model.py
```diff
@@ -185,19 +205,19 @@ def loss(self, theta: torch.Tensor, *context: torch.Tensor):
...
+        Accepts plain device strings (``"cpu"``, ``"cuda"``) as well as
+        rank-qualified CUDA strings (``"cuda:0"``, ``"cuda:1"``, …).  In the
+        latter case ``self.rank`` is set to the integer rank index.
         """
-        if device not in ("cpu", "cuda"):
-            raise ValueError(f"Device should be either cpu or cuda, got {device}.")
+        if "cpu" not in device and "cuda" not in device:
+            raise ValueError(f"Device should contain 'cpu' or 'cuda', got {device}.")
+        if ":" in device:
+            self.rank = int(device.split(":")[1])
```
nihargupte-ph, line 226: Here if the user is doing single GPU training on a multi-gpu (shared) machine and cuda:1 is the device, then this will set self.rank = 1. However, if self.rank=1, then multi-gpu training will be activated. I think instead, there should be an "and" here which sets the rank iff multi gpu is activated otherwise sets to none

dingo/core/posterior_models/base_model.py
```diff
@@ -451,70 +523,179 @@ def train(
...
-        # update loss for history and logging
-        loss_info.update(loss.detach().item(), len(data[0]))
-        loss_info.print_info(batch_idx)
 
-    return loss_info.get_avg()
+        if automatic_mixed_precision:
+            with autocast("cuda"):
+                result = pm.loss(data[0], *data[1:])
+            loss = result[0] if isinstance(result, tuple) else result
+            scaler.scale(loss).backward()
```
nihargupte-ph: Should this be divided by `gradient_updates_per_optimizer_step`? Otherwise suppose we have g1, g2 gradient updates. Then the .backwards pass will contain g1+g2 when really it should be g1+g2/2 if you want an unbiased estimate of the gradient. Also a general question, we aren't really using multiple gradient updates per optimizer step currently right? This is more a for future or testing feature?

dingo/core/utils/trainutils.py
```diff
@@ -96,57 +99,147 @@ def __call__(self, val_loss: float):
...
 
-    def get_avg(self):
+    def _reset_cached_losses(self) -> None:
+        self.cached_losses = []
+        self.cached_n = []
+
+    def update_timer(self, timer_mode: str = "Dataloader") -> None:
+        if self.is_ddp:
+            dt = torch.tensor(time.time() - self.t, device=self.device)
+            dist.barrier()
```
nihargupte-ph: I would think that the only places we nned something sync is during the allreduce for the gradients and the allreduce to compute the loss which goes into the scheduler. I think everywhere else adds overhead right? Or is there a reason there is a .barrier here?

nihargupte-ph (changes requested):
Thanks for the PR! I had a few comments to the code in addition to the more general ones left before. @stephengreen.

docs/source/training_multi_gpu.md
annalena-k: The reason for this is that additional GPU memory is required for syncing the gradients. Since the gradient sync is not necessary with a single GPU, more GPU memory is available compared to multi-GPU training.

dingo/core/utils/trainutils.py
annalena-k: Yes, makes sense.

dingo/core/posterior_models/base_model.py
annalena-k: Yes, makes sense to include this.

dingo/core/posterior_models/base_model.py
annalena-k: I think this is correct. 
We are using multiple gradient updates per optimizer step, at least in the DINGO-T1 paper. This is why I want the feature to be part of the production code as well.

dingo/core/nn/enets.py
```diff
@@ -214,11 +219,13 @@ def __init__(
...
                     activation=activation,
                     dropout_probability=dropout,
-                    use_batch_norm=batch_norm,
+                    use_batch_norm=norm is not None,
                 )
                 for n in range(self.num_res_blocks)
             ]
         )
+        if norm == "LayerNorm":
+            torchutils.replace_BatchNorm_with_LayerNorm(self.blocks)
```
stephengreen: This is a somewhat roundabout way of implementing LayerNorm.  The better solution might be to just reimplement the `MyResidualBlock`, as done [here](https://github.com/dingo-gw/dingo/blob/dingo-t1/dingo/core/nn/resnet.py) on the Dingo-T1 branch. That also maintains consistency with the `dingo-t1` branch.

The interface (`norm` selector) is better than two mutually exclusive options in `dingo-t1`, so it's good to keep that. But it does break some backward compatibility with `dingo-t1` networks.

dingo/core/posterior_models/base_model.py
stephengreen: Did this get added?

dingo/core/utils/backward_compatibility.py
```diff
@@ -140,6 +140,19 @@ def update_model_config(model_settings: dict):
...
+    def _convert_batch_norm(settings: dict):
+        if "batch_norm" in settings:
+            value = settings.pop("batch_norm")
+            settings["norm"] = "BatchNorm" if value in (True, "true", "True") else None
+        for value in settings.values():
+            if isinstance(value, dict):
+                _convert_batch_norm(value)
+
+    _convert_batch_norm(model_settings)
+
```
stephengreen: The easiest thing to do is to instead add a `layer_norm` setting, keeping the `batch_norm` as before. Include a check to make sure they aren't both `True`. This is what's done in the [Dingo-T1 branch](https://github.com/dingo-gw/dingo/tree/dingo-t1) already. Or talk to @annalena-k to make sure you don't run into conflicts.

dingo/core/utils/torchutils.py
```diff
@@ -1,21 +1,146 @@
...
+
+    LayerNorm normalizes each sample independently, so unlike BatchNorm it needs
+    no cross-GPU statistics synchronization under DDP.
+    """
+    for name, child in network.named_children():
+        if isinstance(child, nn.BatchNorm1d):
+            setattr(network, name, nn.LayerNorm(child.num_features, eps=child.eps))
+        else:
+            replace_BatchNorm_with_LayerNorm(child)
+    return network
```
stephengreen: See other comment, this can be removed if instead you re-implement `DenseResidualNet`.

dingo/core/nn/nsf.py
```diff
@@ -136,12 +141,16 @@ def create_base_transform(
...
+            use_batch_norm=norm is not None,
         )
 
     else:
         raise ValueError
 
+    if norm == "LayerNorm":
+        torchutils.replace_BatchNorm_with_LayerNorm(transform)
+    return transform
+
```
stephengreen: For autoregressive flows, the MADE layers impose a certain causality on the feature axis, which the LayerNorm violates. So `LayerNorm` + `rq-autoregressive` will give incorrect results. Raise an error in this situation.

dingo/core/nn/enets.py
stephengreen: Also maintains backwards compatibility.

dingo/core/utils/torchutils.py
stephengreen: Also this implementation uses a different `eps` from `dingo-t1`.

dingo/gw/training/train_pipeline.py
```diff
@@ -437,20 +478,392 @@ def parse_args():
...
+                    import wandb
+
+                    wandb.init(
+                        resume="allow",
+                        dir=train_dir,
+                        **local_settings["wandb"],
+                    )
+                except ImportError:
+                    print("WandB is enabled but not installed.")
+
```
stephengreen, line 642: Even though we'd prefer to use LayerNorm, we should still use SyncBatchNorm in the case of BatchNorm. So restore this line, and add back `replace_BatchNorm_with_SyncBatchNorm()`.
```suggestion

        pm.network = replace_BatchNorm_with_SyncBatchNorm(pm.network)
```

dingo/core/utils/torchutils.py
```diff
@@ -191,51 +316,91 @@ def build_train_and_test_loaders(
...
+            train_dataset, shuffle=True, num_replicas=world_size, rank=rank
+        )
+        test_sampler = DistributedSampler(
+            test_dataset, shuffle=False, num_replicas=world_size, rank=rank
+        )
+        train_loader = DataLoader(
+            train_dataset,
+            batch_size=batch_size,
+            sampler=train_sampler,
+            pin_memory=False,
```
stephengreen: Why is `pin_memory = False` here, but `True` for single-GPU training?

examples/npe_model/train_settings.yaml
```diff
@@ -100,16 +100,18 @@ training:
 local:
   device: cuda # Change this to 'cuda' for training on a GPU.
   num_workers: 32  # num_workers >0 does not work on Mac, see https://stackoverflow.com/questions/64772335/pytorch-w-parallelnative-cpp206
+  # num_gpus: 1  # Set to >1 to enable multi-GPU (DDP) training. batch_size above is the
+                 # total effective batch size; it is divided equally across GPUs.
+                 # When using Condor, request_gpus in the submission file is set automatically.
```
stephengreen: Setting `num_gpus > 1` here is incompatible with `freeze_rb_layer: True` above. (Similar in the other examples.)

dingo/core/utils/torchutils.py
```diff
@@ -1,21 +1,146 @@
...
+    each process draws different random samples.
+    """
+    initial_torch_seed = torch.initial_seed()
+    torch.manual_seed(initial_torch_seed + rank)
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed(initial_torch_seed + rank)
+        torch.backends.cudnn.deterministic = True
+    # NumPy expects seeds in [0, 2**32).
+    reduced_seed = int(initial_torch_seed) % (2**32 - 1)
+    np.random.seed(reduced_seed + rank)
```
stephengreen: Also set the `bilby` random seed here. For `num_workers > 0`, `fix_random_seeds()` will propagate the `torch` seed set here and vary it across workers, which is correct. But for `num_workers = 0`, the `bilby` seed won't get changed, so each GPU will have the same extrinsic parameters.

dingo/gw/training/train_pipeline.py
```diff
@@ -253,56 +229,90 @@ def initialize_stage(
...
         pm.initialize_optimizer_and_scheduler()
 
     # Freeze/unfreeze RB layer if necessary
     if "freeze_rb_layer" in stage:
         if stage["freeze_rb_layer"]:
+            if world_size is not None and world_size > 1:
+                raise ValueError(
+                    "Cannot freeze the RB layer during DDP training — modify "
+                    "the network before wrapping it with DDP."
+                )
```
stephengreen, line 300: This is a bit annoying, and means we can't run the standard DINGO training with the frozen layer on multi-GPU. If we tear down and re-wrap DDP at each stage boundary, we could avoid. Fine for now though.

docs/source/training_multi_gpu.md
```diff
@@ -0,0 +1,216 @@
+# Multi-GPU Training
```
stephengreen, line 1: Needs to be linked from `index.md`, otherwise the new docs don't appear in the toctree.

docs/source/training_multi_gpu.md
```diff
@@ -0,0 +1,216 @@
...
+
+When the `batch_size` is scaled by `num_gpus`, the actual wall-clock speedup is somewhat less than 8× because the 
+gradient all-reduce adds overhead per step (visible as inflated *Time Network* in the log); a factor of 4–6× is typical.
+
+In general, it is recommended to increase the batch size as much as possible to fully utilize the GPU memory.
+
+### Learning rate scaling
+
+When the effective batch size increases by a factor of N, the learning rate should be increased
+by the same factor (linear scaling rule, {cite:p}`Goyal:2017`):
```
stephengreen, line 112: Citation needs to be added to `refs.bib`.

dingo/gw/training/train_pipeline_condor.py
```diff
@@ -30,16 +32,39 @@ def create_submission_file(
...
+        lines.append(f"requirements = {requirements}\n")
+
+    if "num_gpus" in condor_settings:
+        num_gpus = condor_settings["num_gpus"]
+        lines.append(f"request_gpus = {num_gpus}\n")
+        # Cluster-specific full-node templates (MPI-IS).
+        if num_gpus == 8:
+            lines.append("use template : FullNode\n")
+        elif num_gpus >= 6:
+            lines.append(f"use template : FullNode({num_gpus})\n")
```
stephengreen: Is `FullNode` an MPI-IS specific thing? If so, this code could cause a failure for other clusters. I suggest to make this opt-in, for example with a passthrough like `extra_submit_lines:` in the config file, or a flag like `mpi-is: true` (less broadly useful).

dingo/core/nn/enets.py
stephengreen: As it is now, the LayerNorm parameters end up stored under `batch_norm_layers.*` state-dict keys, which also makes it incompatible with dingo-t1.

dingo/gw/training/train_pipeline.py
```diff
@@ -437,20 +478,392 @@ def parse_args():
...
+    -------
+    (complete, resume, epoch) : (bool, bool, int)
+    """
+    if not resume:
+        pm, wfd = prepare_training_new(train_settings, train_dir, local_settings)
+    else:
+        pm, wfd = prepare_training_resume(ckpt_file, local_settings, train_dir)
+
+    global_epoch = Value(ctypes.c_int, pm.epoch)
+    wfd.epoch = global_epoch
```
stephengreen: `global_epoch` is never used except to set the epoch in the `WaveformDataset` here. (It's threaded through various functions though.) It's also not clear to me why we want to store the epoch in the `WaveformDataset`. Can we remove `global_epoch` completely and not set `wfd.epoch`?

dingo/core/posterior_models/base_model.py
```diff
@@ -451,70 +529,183 @@ def train(
...
 
-    return loss_info.get_avg()
+        # Gradients are summed over the accumulated mini-batches, so divide each
+        # loss by the number of accumulation steps to obtain the gradient of the
+        # mean loss over the effective batch (an unbiased estimate that does not
+        # grow with gradient_updates_per_optimizer_step).
+        if automatic_mixed_precision:
+            with autocast("cuda"):
+                result = pm.loss(data[0], *data[1:])
+            loss = result[0] if isinstance(result, tuple) else result
```
stephengreen: Is `result` ever a `tuple`? Is this something for `dingo-t1`?

dingo/core/posterior_models/base_model.py
```diff
@@ -451,70 +529,183 @@ def train(
...
-        dataloader.batch_size,
+        epoch=pm.epoch,
+        len_dataset=_dataset_len(dataloader),
+        batch_size_per_grad_update=effective_bs,
         mode="Train",
         print_freq=1,
+        device=pm.device,
     )
 
+    scaler = GradScaler("cuda") if automatic_mixed_precision else None
```
stephengreen: This is recreated every epoch. Can we persist it (create it in `train()` and pass it)? At present, it loses its calibrated state each epoch and we would lose the first few iterations per epoch due to overflow (according to Claude).

dingo/core/posterior_models/base_model.py
```diff
@@ -451,70 +529,183 @@ def train(
...
-    return loss_info.get_avg()
+        # Gradients are summed over the accumulated mini-batches, so divide each
+        # loss by the number of accumulation steps to obtain the gradient of the
+        # mean loss over the effective batch (an unbiased estimate that does not
+        # grow with gradient_updates_per_optimizer_step).
+        if automatic_mixed_precision:
+            with autocast("cuda"):
+                result = pm.loss(data[0], *data[1:])
+            loss = result[0] if isinstance(result, tuple) else result
+            scaler.scale(loss / gradient_updates_per_optimizer_step).backward()
```
stephengreen: With `gradient_updates_per_optimizer_step > 1` and DDP, this will all-reduce even on the accumulation-only steps, which is wasteful. The gradients only need to be synchronized on the final backward pass. This can be handled using `DDP.no_sync()` context, e.g.,
```python
  from contextlib import nullcontext                                                            
                                                                                                
  is_step_batch = (batch_idx + 1) % gradient_updates_per_optimizer_step == 0                    
  sync_ctx = (                                                                                  
      pm.network.no_sync()                                                                      
      if isinstance(pm.network, DDP) and not is_step_batch                                      
      else nullcontext()                                                                        
  )                                                                                             
  with sync_ctx:                                                                                
      loss = pm.loss(data[0], *data[1:])                                                        
      (loss / gradient_updates_per_optimizer_step).backward()                                   
  if is_step_batch:                                                                             
      optimizer.step()
```

dingo/gw/training/train_pipeline.py
stephengreen: Perhaps give a warning saying it's better to use LayerNorm.

docs/source/training_multi_gpu.md
```diff
@@ -0,0 +1,216 @@
...
+    batch_size: 4096
+    optimizer:
+      lr: 1.0e-5
+```
+
+```yaml
+# 8-GPU DDP — scale batch_size and lr by num_gpus
+local:
+  device: cuda
+  num_workers: 32   # scale with num_gpus, e.g. 4–8 per GPU
```
stephengreen: Isn't this the per-GPU number of workers (so this request would give 8*32=256 workers)? If we're specifying the total batch_size in the config file, we could alternatively divide `num_workers` by the number of GPUs within the code.

dingo/gw/training/train_pipeline.py
```diff
@@ -437,20 +478,392 @@ def parse_args():
...
+        batch_size = train_settings["training"]["stage_0"]["batch_size"]
+        print("\nBuilding SVD for initialization of embedding network.")
+        initial_weights["V_rb_list"] = build_svd_for_embedding_network(
+            wfd=wfd,
+            data_settings=train_settings["data"],
+            asd_dataset_path=train_settings["training"]["stage_0"]["asd_dataset_path"],
+            batch_size=batch_size,
+            out_dir=train_dir,
+            **model_kwargs["embedding_kwargs"]["svd"],
+        )
```
stephengreen, line 847: This lost the line
```python
num_workers=local_settings["num_workers"]
```
so it will run with `num_workers` 0. Is that intentional?

dingo/core/utils/torchutils.py
```diff
@@ -1,21 +1,146 @@
...
+def set_seed_based_on_rank(rank: int) -> None:
+    """
+    Set NumPy and Torch seeds for a DDP worker process based on *rank* so that
+    each process draws different random samples.
+    """
+    initial_torch_seed = torch.initial_seed()
+    torch.manual_seed(initial_torch_seed + rank)
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed(initial_torch_seed + rank)
+        torch.backends.cudnn.deterministic = True
```
stephengreen: What's the reason for this line? Claude says that it could reduce performance as it restricts to bitwise-reproducible algorithms which aren't really needed unless you want run-to-run reproducibility. It's also only applied to the DDP path, not single-GPU training. Also apparently it only affects convolutions. Suggest dropping it.

dingo/core/posterior_models/base_model.py
```diff
@@ -451,70 +529,183 @@ def train(
...
-        loss_info.update(loss.detach().item(), len(data[0]))
-        loss_info.print_info(batch_idx)
 
-    return loss_info.get_avg()
+        # Gradients are summed over the accumulated mini-batches, so divide each
+        # loss by the number of accumulation steps to obtain the gradient of the
+        # mean loss over the effective batch (an unbiased estimate that does not
+        # grow with gradient_updates_per_optimizer_step).
+        if automatic_mixed_precision:
+            with autocast("cuda"):
```
stephengreen: Validate `"cuda"` device here, and in the `GradScaler("cuda")` line above, in cases where one might be using CPU.

dingo/core/utils/torchutils.py
```diff
@@ -1,21 +1,146 @@
...
+    # NumPy expects seeds in [0, 2**32).
+    reduced_seed = int(initial_torch_seed) % (2**32 - 1)
+    np.random.seed(reduced_seed + rank)
+
+
+def setup_ddp(rank: int, world_size: int, port: int = 12355) -> None:
```
stephengreen, line 76: Allow the `port` to be settable from `local:` config, so no collisions occur on a node.

tests/core/test_multi_gpu.py
```diff
@@ -0,0 +1,403 @@
...
+def free_port():
+    """Find a free port for the gloo process group."""
+    import socket
+
+    with socket.socket() as s:
+        s.bind(("", 0))
+        return s.getsockname()[1]
+
+
+class TestGlooDDP:
```
stephengreen, line 371: When a distributed test regresses, it hangs rather than fails.

Suggest two changes:
1. In `_setup_gloo`, pass `timeout=timedelta(seconds=60)` to `init_process_group`.
2. The spawn/join boilerplate is repeated in all three tests — factor it into a helper with `daemon=True`, a join timeout, and terminate-on-failure:

```python
def _run_workers(target, world_size, port, result_queue):
    """Run one worker per rank; fail fast if any hangs or crashes."""
    processes = [
        mp.Process(
            target=target,
            args=(rank, world_size, port, result_queue),
            daemon=True,
        )
        for rank in range(world_size)
    ]
    try:
        for p in processes:
            p.start()
        for p in processes:
            p.join(timeout=60)
        for p in processes:
            assert not p.is_alive(), "Distributed worker hung."
            assert p.exitcode == 0, f"Worker failed with code {p.exitcode}"
    finally:
        for p in processes:
            if p.is_alive():
                p.terminate()
```

stephengreen (changes requested):
Very nice PR! I found a bunch of small changes, but then I'm happy with it. Thanks a lot for all this careful work!

dingo/core/utils/torchutils.py
nihargupte-ph: Added a new function `torchutils.set_seed_based_on_rank(rank: int) -> None:` which will set the random seed including the bilby random seed.

dingo/gw/training/train_pipeline.py
nihargupte-ph: Done and added back the SyncBatchNorm. In principle, you could still keep BatchNorm and remove SyncBatchNorm completely as long as the batches are big enough that the batch statistics do not drift. However, having a SyncBatchNorm is the safer option. After all, we would use Layer norm if we want speed.

dingo/core/utils/torchutils.py
nihargupte-ph: I am not sure actually, I think this is more a question for @annalena-k. For now I reset it back to `pin_memory=True`

dingo/core/nn/nsf.py
nihargupte-ph: fixed.

dingo/core/nn/enets.py
nihargupte-ph: True, I didn't realize that's how it was implemented in the `dingo-t1` branch. I've copied that code over to this branch including the `layer_norm_layers.*` syntax. In backward_compatibilty added a function which will convert old network settings which have batch_norm: True to norm: BatchNorm

dingo/gw/training/train_pipeline_condor.py
nihargupte-ph: Yes, I think it's a MPI-IS only thing. Added it an option to add via extra-lines.

docs/source/training_multi_gpu.md
nihargupte-ph: Done

dingo/core/posterior_models/base_model.py
nihargupte-ph: I am not sure, removing for now but if there is an error when merging the T1 branch we could consider adding it back in.

dingo/gw/training/train_pipeline.py
nihargupte-ph: True, I removed it and now grepping the codebase I don't see any instances of wfd.epoch.

dingo/core/posterior_models/base_model.py
nihargupte-ph: Good catch, I agree.

dingo/gw/training/train_pipeline.py
nihargupte-ph: I believe this has to do with https://github.com/dingo-gw/dingo/pull/349

docs/source/training_multi_gpu.md
nihargupte-ph: Done, a new variable defined in `train_pipeline` called `num_workers_per_gpu`.

dingo/core/posterior_models/base_model.py
nihargupte-ph: Yes, now saved before the training loop and passed as an agrument to `train_epoch`.

dingo/core/utils/torchutils.py
nihargupte-ph: Dropped.

docs/source/training_multi_gpu.md
nihargupte-ph: done

dingo/core/utils/torchutils.py
nihargupte-ph: Done.

dingo/gw/training/train_pipeline.py
nihargupte-ph: I believe this has to do with https://github.com/dingo-gw/dingo/pull/349

dingo/core/utils/torchutils.py
nihargupte-ph: Dropped.

dingo/core/utils/torchutils.py
nihargupte-ph: Done.

dingo/core/utils/torchutils.py
nihargupte-ph: Done

tests/core/test_multi_gpu.py
nihargupte-ph: done

stephengreen:
Thanks @nihargupte-ph, I went through all the changes in 83ce8aa9 and everything checks out. I've resolved all the review threads. (One of my comments was wrong: the SVD loader already pinned `num_workers=0` on `main` after #349, so the documentation you added is the right outcome.)

I also checked the new `DenseResidualNet` layout against the published Dingo-T1 checkpoint: its coupling conditioners have `initial_layer.weight` of shape `(512, 7)` and all norm parameters under `layer_norm_layers.*`, so context enters only through the per-block GLU with no initial-layer concat. That matches exactly what this branch builds, so T1 networks will load.

To save a round trip I pushed one small commit on top (0e81d12d):

- **Validation now runs under autocast when `automatic_mixed_precision` is on.** `test_epoch` gained the flag (passed from both call sites in `train()`), and the loss is evaluated inside `autocast(<device type>)`. Previously validation ran in fp32 while training ran in fp16, which made validation several times slower per batch and computed the test loss at a different precision from the train loss. This is the second hunk of @annalena-k's fix (dingo-development `1761755d`); the first hunk (the B200 `set_fastpath_enabled(False)`) is transformer-specific and already lives on the `transformer` branch. Test added in `test_amp.py`.
- **`MyResidualBlock` applies the context GLU along `dim=-1` instead of `dim=1`.** Identical for 2D input, but it lets token-batched `[batch, tokens, features]` inputs be gated per token, which is what the `transformer` branch needs. Regression test ported from that branch. This removes the only semantic difference between the two branches' `resnet.py`, so the later merge is purely a signature reconciliation.
- **Removed `tests/gw/mm.png`**, which came along in 83ce8aa9 by accident.

Two requests before merging:

1. **Please squash-merge.** The image is still in the branch history, and the history also contains a couple of try/revert pairs (6221e3e9/3c5d38ae, 84fb96b4/6f48ffcb).
2. **Refresh the PR description**, which predates the norm work: `norm: LayerNorm` / the `BatchNorm` → `SyncBatchNorm` warning, `ddp_port`, `condor.extra_submit_lines`, `no_sync()` during accumulation, and `num_workers` being the total across GPUs.

Approving now; the two items above don't need another review round.

stephengreen (approved):
All review comments addressed and verified locally (58 tests pass, CI green). See my comment for the one follow-up commit I pushed and the squash-merge request.

nihargupte-ph (approved):

PR #368, opened 2026-03-02
Implement pre-commit hooks: black and isort

annalena-k:
This PR implements automatic formatting via `black` and import sorting via `isort` using added **pre-commit hooks**, so formatting and import ordering are checked on every commit rather than caught in review.                                                                                                                  
                                                            
**Changes:**
  - `.pre-commit-config.yaml` — new file; runs `black` (25.1.0) and `sort` (5.13.2) as pre-commit hooks
  - `pyproject.toml` — adds `[tool.black]` (line-length 88, py310/py311) and `[tool.isort] `(black-compatible profile, dingo as first-party) config sections; adds `isort` and `pre-commit` to dev dependencies
  - `.github/workflows/pytest.yml` — adds a lint CI job that runs `black --check` and `isort --check --diff`, blocking PRs that aren't compliant

All files had to be reformatted such that the new CI test passes.

To set this up locally, developers have to run the following once:
```
  uv sync --group dev   # or: pip install .[dev]
  pre-commit install
```
The above instructions were added to the documentation. After initializing pre-commit, every commit looks like this: 
image
If e.g. `isort` fails, the commit is rejected and `isort` is automatically applied to the files where it failed. The modified files can be added to git and committed again.

Note: When the pre-commit hook is activated, it only runs on files which are staged (i.e. files with changes to be committed). This means that if you merged in updates from another branch (e.g. main) which haven't been formatted and then only files with merge conflicts get staged in the next commit, the hook will format the modified files, but non-staged files from the merge might still need the formatting to be applied. If this is not done manually, the CI test will fail, despite the pre-commit hooks passing locally.


vincentberenz (approved):
Great to streamline formatting !
I added a minor comment

ewencedr:
I think it would be a good idea to use Ruff instead, since it combines Black and isort
in one tool and is much faster. On top of that, we'd also get linting.

I already did this in my fork, and it also includes the merge with main. So if you think
it's a good idea, we could just move this branch to that state. That's probably cleaner
than opening a new PR, either against this branch or a competing one to main.

Fork: https://github.com/ewencedr/dingo/tree/ruff-pre-commit

PR #376, opened 2026-04-08
Remove getenv=True.

AlexandreGoettel:
Since yesterday, setting `getenv=True` in an htcondor job on CIT will automatically fail the job. This PR is an attempt to fix this by explicitely setting getenv to `HOME` when needed of `False` otherwise.


annalena-k:
If I remember correctly, the standard (non-CIT) htcondor cluster I work on actually requires `getenv=True`.
Can you please modify the PR such that these changes are only made when `simple-submission=False`?
When `simple-submission=True`, the original code should be run since this mode expects no CIT-like infrastructure.

PR #379, opened 2026-04-22
Allow resuming condor training without train_settings.yaml

annalena-k:
## Problem
**Setting and expected behavior:**
I wanted to fine-tune DINGO models by appending a new training stage and continuing training in a different folder to separate the two runs clearly. After appending the stage, I expected that I would only need the modified model checkpoint and the `local_settings.yaml` file to start the fine-tuning training since the model metadata should contain all information that would in principle be part of the `train_settings.yaml` file.

**Actual behavior:**
However, when I tried to start the training jobs with condor, the code required `train_settings.yaml` to be present even though all necessary settings are already available in `local_settings.yaml` and the checkpoint. 
Additionally, I didn't copy over the `wandb` folder which resulted in another failure since wandb's `resume="must"` mode requires the local run state to exist.

## Fixes

- `train_pipeline_condor.py`: Read condor settings from `local_settings.yaml` when available; only fall back to `train_settings.yaml` if `local_settings.yaml` does not exist (i.e. on first submission).
- `train_pipeline.py`: Switch wandb `resume="must"` to `resume="allow"`, which resumes if local state is present and otherwise re-initializes from the server.

### DIFF
--- a/dingo/gw/training/train_pipeline.py
+++ b/dingo/gw/training/train_pipeline.py
@@ -218,7 +218,7 @@ def prepare_training_resume(
             import wandb
 
             wandb.init(
-                resume="must",
+                resume="allow",
                 dir=train_dir,
                 **local_settings["wandb"],
             )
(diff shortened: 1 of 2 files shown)


nihargupte-ph (approved):
LGTM

PR #381, opened 2026-05-04
Optimize synthetic phase reconstruction

lorenzopompili00:
This MR adds a new option `use_dft_phase_decomposition` to `WaveformGenerator` and   `NewInterfaceWaveformGenerator` that enables an optimized synthetic phase reconstruction (plus a vectorization of the phase-grid likelihood it feeds into). I also attach a short [note](https://github.com/user-attachments/files/27376788/dft_phase_decomposition.pdf) explaining the method in more detail. 
 
Instead of evaluating each $(\ell, m)$ mode in the inertial frame and reassembling, the polarizations are evaluated at $N = 2\ell_{\max}+1$ equally-spaced $\phi_c$ offsets and inverted with a DFT to recover the $\tilde h_{+,m}(f)$ components.  

* For SEOBNRv5PHM this calls a new `generate_multi_phase_fd_polarizations` method in pyseobnr (see https://git.ligo.org/waveforms/software/pyseobnr/-/merge_requests/132), which evaluates the dynamics and co-precessing modes once and re-projects at each $\phi_c$. 
* For IMRPhenomXPHM the $N$ polarizations come from $N$ independent calls to the polarization routine (which uses the cheap direct projection of Appendix C of https://arxiv.org/abs/2004.06503). This is not optimal (one could also cache the rotations and co-precessing modes) but still slightly faster than the default path. For IMRPhenomXPHM it is also more faithful to the model than the individual-mode path, which computes the orbital angular momentum magnitude using a slightly different prescription than the model, see e.g., https://github.com/dingo-gw/dingo/blob/8094b56d0c6fbe01d01c1b4b0d40b9cb576fb807/tests/gw/waveform_generator/test_wfg_m.py#L111-L115

I have tested this re-running importance sampling for some O4b events, see the figures in the comment below.

One could speed this up further by caching the polarizations for both the synthetic phase and importance sampling, but would require more substantial code changes.


lorenzopompili00:
Sharing a few extra diagnostic plots, comparing IS between ProdPE and this PR for a few O4b events with good efficiency.

validation_summary_S240527en
validation_summary_S240618ah
validation_summary_S241125n
validation_summary_S250101k
validation_summary_S250118t

dingo/gw/likelihood.py
```diff
@@ -438,22 +438,24 @@ def _log_likelihood_phase_grid_mode_decomposed(self, theta, phases=None):
...
+        cs = np.array(list(rho2opt_crossterms.values()))
+        rho2opt = rho2opt_const + (
+            cs[:, None] * np.exp(-1j * deltas[:, None] * phases_arr[None, :])
+        ).real.sum(axis=0)
+
+        m_arr = np.array(m_vals)
+        k_arr = np.array([kappa2_modes[m] for m in m_vals])
+        kappa2 = (
+            k_arr[:, None] * np.exp(-1j * m_arr[:, None] * phases_arr[None, :])
+        ).real.sum(axis=0)
```
nihargupte-ph, line 466:
```suggestion
        # Vectorised evaluation over the phase grid. Per phase ph:
        #   rho2opt(ph) = rho2opt_const
        #       + sum_{(m,n)} (crossterm_{m,n} * exp(-i*(n-m)*ph)).real
        #   kappa2(ph)  = sum_m (kappa2_modes[m] * exp(-i*m*ph)).real
        #
        # The cross terms only depend on the mode difference delta = n - m, so we
        # accumulate them by delta before touching the grid. For m in -4..4 this
        # collapses 36 pairs onto 8 distinct deltas, i.e. a 4.5x smaller
        # exponential matrix for identical arithmetic.
        phases_arr = np.asarray(phases)

        crossterms_by_delta = defaultdict(complex)
        for (m, n), c in rho2opt_crossterms.items():
            crossterms_by_delta[n - m] += c
        deltas = np.array(sorted(crossterms_by_delta))
        cs = np.array([crossterms_by_delta[delta] for delta in deltas])
        rho2opt = rho2opt_const + (
            cs[:, None] * np.exp(-1j * deltas[:, None] * phases_arr[None, :])
        ).real.sum(axis=0)

        m_arr = np.array(m_vals)
        k_arr = np.array([kappa2_modes[m] for m in m_vals])
        kappa2 = (
            k_arr[:, None] * np.exp(-1j * m_arr[:, None] * phases_arr[None, :])
        ).real.sum(axis=0)
```

The cross terms are still expanded over all 36 (m, n) pairs. They only depend on
`delta = n - m`, so for m in -4..4 they collapse onto 8 distinct deltas: a 4.5x smaller
exponential matrix for identical arithmetic. `sorted()` isn't needed for correctness but
keeps `deltas`/`cs` reproducible rather than dict-insertion-ordered. `kappa2` is
unchanged from what's on the branch. 

The overall speedup is 1.1x only though.

nihargupte-ph (approved):
Looks good to me. I suggest a small optimization, but I think the bulk has already been optimized quite a lot. Great work!

dingo/gw/likelihood.py
nihargupte-ph: Note this needs `from collections import defaultdict` at the top of the module

lorenzopompili00: Thanks! I just added the optimization you suggested.

lorenzopompili00:
Thanks for the review @nihargupte-ph! Before merging, I wanted to raise a couple of small points, I would welcome input from others too:
- Is `use_dft_phase_decomposition` a good name for the option, or would something else be clearer?
- It currently defaults to `use_dft_phase_decomposition: bool = False`. Should the default be `True` instead?

nihargupte-ph:
> Thanks for the review @nihargupte-ph! Before merging, I wanted to raise a couple of small points, I would welcome input from others too:
> 
> * Is `use_dft_phase_decomposition` a good name for the option, or would something else be clearer?
> * It currently defaults to `use_dft_phase_decomposition: bool = False`. Should the default be `True` instead?

Maybe `obtain_modes_with_dft_of_polarizations`? 
I think it should default to True, it is exact after all.

lorenzopompili00:
Thanks @nihargupte-ph!

On the naming, I'd rather avoid "modes": usually by mode we mean an (l, m) multipole, while this method returns the m-decomposition of the polarizations, and its whole point is that it never builds the individual modes. For now I'll keep `use_dft_phase_decomposition` unless someone has a better suggestion.

I changed the default to True. This caught a couple small bugs in the `MultibandedFrequencyDomain` case, which was previously untested with the DFT path (missing truncation on domains where `f_max / delta_f` is not a power of two, and a missing DFT option in the MFD branch which are now fixed).

Also, to avoid breaking compatibility, if the phase grid cannot be sized (no `mode_list`, no `DEFAULT_ELL_MAX` entry), `generate_hplus_hcross_m` warns and falls back to the individual-mode path instead of raising an error. The flag can also be overridden per run in the standard way, e.g. `importance-sampling-settings = {synthetic_phase: {..., use_dft_phase_decomposition: False}}`.

PR #382, opened 2026-05-11
Fix TypeError when model_init is None in data_generation and sampling

annalena-k:
## Problem

Calling `resolve_filename_with_transfer_fallback(args.model_init)` with `args.model_init=None` raises a `TypeError`:

```
File "dingo/pipe/data_generation.py", line 34, in __init__
    self.model_init = resolve_filename_with_transfer_fallback(args.model_init) or args.model_init
File "bilby_pipe/utils.py", line 900, in resolve_filename_with_transfer_fallback
    if os.path.isfile(filename):
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

`--model-init` is an optional argument (only required for GNPE models) that defaults to `None`. The bug was introduced in #372, which added `resolve_filename_with_transfer_fallback` calls for `model_init` without guarding against the case where `model_init` is not set.

## Fix

Guard the call with `args.model_init and (...)` so that `None` short-circuits before `resolve_filename_with_transfer_fallback` is reached. The same fix is applied symmetrically in both `data_generation.py` and `sampling.py`.

### DIFF
--- a/dingo/pipe/data_generation.py
+++ b/dingo/pipe/data_generation.py
@@ -31,7 +31,9 @@ logger.name = "dingo_pipe"
 class DataGenerationInput(BilbyDataGenerationInput):
     def __init__(self, args, unknown_args, create_data=True):
         self.model = resolve_filename_with_transfer_fallback(args.model) or args.model
-        self.model_init = resolve_filename_with_transfer_fallback(args.model_init) or args.model_init
+        self.model_init = args.model_init and (
+            resolve_filename_with_transfer_fallback(args.model_init) or args.model_init
+        )
 
         Input.__init__(self, args, unknown_args)
         # Generic initialisation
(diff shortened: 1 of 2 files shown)


dingo/pipe/data_generation.py
```diff
@@ -31,7 +31,9 @@
 class DataGenerationInput(BilbyDataGenerationInput):
     def __init__(self, args, unknown_args, create_data=True):
         self.model = resolve_filename_with_transfer_fallback(args.model) or args.model
-        self.model_init = resolve_filename_with_transfer_fallback(args.model_init) or args.model_init
+        self.model_init = args.model_init and (
+            resolve_filename_with_transfer_fallback(args.model_init) or args.model_init
+        )
```
nihargupte-ph, line 36: I guess the idea to keep the "or" here is if 

1). The model_init is not None
2). You do `resolve_filename_with_transfer_fallback` but this gives None because it's not filepath 
3). It default to the name (which is not a file). 

That last "or" seems kind of uneccesary to me.  

In retrospect, maybe using `args.model_init and resolve_filename_with_transfer_fallback(args.model_init)` is better. This would also be for the main model.

annalena-k: Thanks for the comment!

I understand it a little differently: 
The `or` in this line is not reached when `args.model_init = None` due to Python's short-circuit evaluation: if `args.model_init` is None (falsy), the `and` returns `None` immediately without ever calling the part behind the `and`. Therefore, (1.) isn't relevant for the `or`.

So, currently, the code is doing:
  - `args.model_init = None` => None
  - `args.model_init` is a valid path and the file exists => resolved path from `bilby_pipe`
  - `args.model_init` is a string but the file is not found => continue with `args.model_init` as string (i.e. no changes)
 
Does this mean that you would remove the part after the `or` (i.e. `or args.model_init`)?
In this case, the code would have `args.model_init = None` although a string was provided as the init network. What is the reason to remove the fallback to a string? 

If you want it to be removed, I'd suggest raising an error instead that points out that the init network file cannot be found.

What do you think? :)

nihargupte-ph (approved):

PR #383, opened 2026-06-05
LVK fixes for 0.9.8

AlexandreGoettel:
This PR implements four changes, one of which is necessary for DINGO to be used in (offline) production. In order of importance:

1. Correct the sample() function in asimov.py to be compatible with asimov >= 0.5 (this is absolutely necessary)
2. Modify the formatting of the file given to PESummary so it can run (also necessary but could technically be fixed on the PESummary side instead)
3. Change the asimov/dingo.ini file to allow for importance sampling updates, and to correctly pass the condor user (I would not make a new release for this, but since we need a release anyway, this will make things easier)
4. Change  recursive_check_dicts_are_equal to instead use np.allclose for floats. This one is not needed if we run everything on Hypatia, but is required to run IS jobs on the broader OSG. @stephengreen suggested that instead, a new function called recursive_check_dicts_are_allclose be created, which I plan to implement

### DIFF
--- a/dingo/asimov/asimov.py
+++ b/dingo/asimov/asimov.py
@@ -150,10 +150,10 @@ class Dingo(Pipeline):
         result_files = glob.glob(
             os.path.join(rundir, "result", f"*importance_sampling.hdf5")
         )
-    if len(result_files) == 0: 
-        raise ValueError("Importance sampling result file not found")
-    elif len(result_files) > 1: 
-        raise ValueError("multiple importance sampling result files found")
+        if len(result_files) == 0: 
+            raise ValueError("Importance sampling result file not found")
+        elif len(result_files) > 1: 
+            raise ValueError("Multiple importance sampling result files found")
 
         # pesummary can't presently read a result file containing MultibandedFrequencyDomain
         # This is a hotfix to create a copy of the result file with UniformFrequencyDomain
--- a/dingo/asimov/asimov.py
+++ b/dingo/asimov/asimov.py
@@ -18,6 +18,8 @@ from asimov.pipeline import (
     PESummaryPipeline,
 )
 
+from dingo.gw.result import Result
+
 
 class Dingo(Pipeline):
     """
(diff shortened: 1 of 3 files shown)


dingo/asimov/asimov.py
```diff
@@ -144,16 +137,37 @@ def build_dag(self, psds=None, user=None, clobber_psd=False, dryrun=False):
...
+        if absolute:
+            rundir = os.path.abspath(self.production.rundir)
+        else:
+            rundir = self.production.rundir
+        self.logger.info(f"Rundir for samples: {rundir}")
+        result_files = glob.glob(
+            os.path.join(rundir, "result", f"*importance_sampling.hdf5")
         )
-        return os.path.join(results_dir, results_filenames[0])
+        assert len(result_files) == 1
```
nihargupte-ph: I think probably better to raise an exception here for interpretability.

```suggestion
    if len(result_files) == 0: 
        raise ValueError("Importance sampling result file not found")
    elif len(result_files) > 1: 
        raise ValueError("multiple importance sampling result files found")
```

This also assumes that pesummary will always only take IS results. I guess that is fine for now since that's what we'll be doing in prod with asimov.

dingo/asimov/dingo.ini
```diff
@@ -95,13 +95,16 @@ maximum-frequency={
...
+importance-sampling-updates={{ data['importance sampling updates'] }}
+{%- endif %}
 
 ################################################################################
 ##  Job submission arguments
 ################################################################################
 
 accounting = {{ scheduler["accounting group"] }}
-accounting-user = {{ scheduler["accounting user"] | default: None }}
+accounting-user = {{ config['condor']['user'] }}
```
nihargupte-ph, line 107: I guess this is a change coming from the latest asimov blueprints?

dingo/asimov/asimov.py
```diff
@@ -144,16 +137,37 @@ def build_dag(self, psds=None, user=None, clobber_psd=False, dryrun=False):
...
+
+        # pesummary can't presently read a result file containing MultibandedFrequencyDomain
```
nihargupte-ph, line 158: This surprises me a bit. What error did you get when loading the multibanded frequency domain file? I say that because we used multibanded networks and pesummary for the GWTC5 events w/o issues, but we didn't use asimov+pesummary so maybe there is a cleaner way.

nihargupte-ph (changes requested):

nihargupte-ph:
Thanks for the timely PR @AlexandreGoettel ! I left a few comments.

dingo/asimov/dingo.ini
AlexandreGoettel: Kinda, we changed this w.r.t. GWTC-5 because in general it's not "good practice" to have usernames in blueprints.

dingo/asimov/asimov.py
AlexandreGoettel: The problem is [this line](https://git.ligo.org/lscsoft/pesummary/-/blob/master/pesummary/gw/file/formats/dingo.py?ref_type=heads#L124) which reads delta_f from the MultibandedFrequencyDomain and receives an array (instead of a float with UniformFrequencyDomain). This causes [this if statement](https://git.ligo.org/lscsoft/pesummary/-/blob/master/pesummary/gw/conversions/snr.py?ref_type=heads#L169) in get_precessing_snr to crash. This also happens in get_multipole_snr. So, if you were using pesummary without specifically specifying either calculate-precessing-snr or calculate-multipole-snr, you would not have gotten any error

AlexandreGoettel: I'll note that the best long term solution is to change this line on the pesummary side. I just think this hotfix on the dingo side is better for the production timeline. On the other hand, since we have to wait for 1.7 release anyway, it might be worth opening a PR on the pesummary side ASAP

nihargupte-ph (approved):

PR #385, opened 2026-06-30
first repo-wide CLAUDE.md file

hcampe:
We have created a first dingo-specific CLAUDE.md file and added instructions to the README for how to use it along with user-specific commands. We have compared several candidates for this file by letting claude execute a certain task with each of them and then comparing the results. 

Probably the best option would be that everyone uses it this week and modifies it where necessary s.t. at the end of the week we have a good version that can be included into the repo.


nihargupte-ph:
LGTM

stephengreen:
image

PR #387, opened 2026-07-01
Add unit tests across core, gw, and pipe (+ two small fixes)

lorenzopompili00:
Substantially expands the test suite — from 101 to 314 passing tests — focused on previously-untested logic, with cheap fixtures so it stays fast.

Source fixes (both surfaced by the new tests, under numpy>2.0):
- `trainutils.py`: `np.Inf` -> `np.inf`
- `gw/result.py::sample_calibration_parameters`: cast numpy-scalar prior params to native Python before `repr()`, so the calibration path round-trips through bilby (works around an upstream bilby `Prior.__repr__` bug; PR proposed upstream https://github.com/bilby-dev/bilby/pull/1108).

### DIFF
--- a/dingo/core/utils/trainutils.py
+++ b/dingo/core/utils/trainutils.py
@@ -56,7 +56,7 @@ class EarlyStopping:
         self.counter = 0
         self.best_score = None
         self.early_stop = False
-        self.val_loss_min = np.Inf
+        self.val_loss_min = np.inf
         self.delta = delta
         if metric not in ["training", "validation"]:
             raise ValueError(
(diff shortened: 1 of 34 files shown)


nihargupte-ph (commented):
Overall a good PR thanks Lorenzo. The only issue I have is that most of the tests are related to typing or checking inputs and outputs are reasonable. But many of these would be trivially satisfied. I guess that's fine though, that is the point of a test. In the future, it will be better for people to write tests as they write functions, as it is hard to come up with tests after the fact.

nihargupte-ph (changes requested):
I think we should wait for the changes to the sampler befor merging. We can pass the sampler branch to claude and ask it to update the tests.

Issue #152, opened 2023-03-02
Improved logging


stephengreen:
We should use the `logging` package rather than `print` statements throughout the code, issuing either warnings or info depending on circumstances. This will require setting up the logger and changing all print statements. Although this could be automated, there will be some work involved in choosing between warning and info for each one.

For dingo_pipe we do use a logger, but this is the logger created for bilby_pipe, so everything says "BILBY_PIPE". This should be changed as well.

filippo-santoliquido:
Hi! I wanted to raise the idea of setting up a logger in Dingo, as I think it would be quite useful. Following @mpuerrer's comments, it makes sense to include the Bilby logger, since Dingo relies on Bilby and the two loggers can be combined — resulting in both `Bilby - INFO` and `Dingo - INFO` entries in the same log file.

From a technical standpoint, the approach would be to import logging, set up the Bilby logger using [`bilby.core.utils.log.setup_logger`](https://lscsoft.docs.ligo.org/bilby/api/bilby.core.utils.log.setup_logger.html), and replace any `print()` calls with the appropriate logger calls (e.g. `logger.info()`). For this to work correctly, Dingo would need its own named logger (e.g. `logging.getLogger("dingo")`) sharing the same file handler as the Bilby logger — either by explicitly adding it or by configuring it at the root logger level — so that both loggers write to the same output file.

PR #388, opened 2026-07-01
 Improved logging (issue #152)

cmfabbri:
Replaces print() statements with logging calls, and integrates the dingo and bilby loggers so that both write to the same log file.

**Changes:**
- dingo/core/utils/logging_utils.py - updated; sets up a "dingo" logger that shares bilby's file handler when bilby is available, falling back to a standalone file handler otherwise
- replaced print() calls with logger.info() or logger.warning() 
- files in dingo/pipe/ now use logging.getLogger("dingo.pipe") instead of bilby_pipe's logger
-  print() kept in ls_cli.py, asimov.py (dryrun blocks) and __main__ debug blocks

### DIFF
--- a/dingo/core/dataset.py
+++ b/dingo/core/dataset.py
@@ -4,6 +4,7 @@ import h5py
 import numpy as np
 import pandas as pd
 
+from dingo.core.utils.logging_utils import logger
 from dingo.core.utils.misc import get_version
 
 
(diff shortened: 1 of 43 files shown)


stephengreen (commented):
You should make sure that the creation of the logger is consistent with what @hcampe is doing with hydra.

PR #389, opened 2026-07-01
Factorized sampler: a composable Step-based inference spine

stephengreen:
## Summary

This PR replaces the monolithic `Sampler`/`GWSampler`/`GWSamplerGNPE` classes with a factorized inference spine: the posterior is sampled by an explicit **chain of steps** — factors (networks, pins, tables), reparametrizations, and corrections — composed by a `ChainComposer` over a per-event `GWSamplerContext`. Every existing workflow (NPE, multi-iteration GNPE with density recovery, importance sampling with synthetic phase) runs on the spine bit-exactly, and the spine's first new physics capability — DINGO-BNS single-step chirp-mass GNPE (Dax et al., Nature 639, 49 (2025)) — runs end-to-end through `dingo_pipe` on real GW170817 data.

## Architecture

- **`dingo/core/factors.py`** — the domain-agnostic spine: `Factor` / `Reparametrization` / `TargetCorrection` step kinds (`FlowFactor`, `DeltaFactor`, `SampleTableFactor`, `ProxyOffsetReparam`, `GibbsBlock`), `Stage`/`ChainComposer` (fan-out, point-mass and 1:1 multiplicity rules, kind-aware reverse-fold `log_prob`), and the `ComposedSampler` base.
- **`dingo/gw/inference/`** — split by responsibility: `context.py` (`GWSamplerContext`: event data with its network-input, prior, and likelihood views; importance-sampling representations via immutable `derive()`), `steps.py` (GNPE factors, `SyntheticPhaseFactor`, the `RAToEventFrame`/`RAToTrainingFrame` pair, `SpinConventionReparam`), `sampler.py` (`GWComposedSampler` + chain builders), `scan.py` (the chirp-mass scan).
- **`Result`** carries no construction knowledge: prior, domain, and likelihood delegate to the context (reconstructed from metadata for from-file results); synthetic phase runs as a chain rooted in the proposal sample table.
- Results record **structured provenance**: `settings["sampler"]` holds the serialized chain description, model paths, and (where applicable) density-recovery and scan recipes.

## Binary neutron stars

- Heterodyning transforms (`HeterodynePhase`, batch-native `factor_fiducial_waveform`) ported from #355; `d_inner_h` phase decomposition (`return_rho2opt`) ported from #354; matched-filter SNR reporting (`return_aux_snr`) added for the scan.
- **Fixed-proxy chain** `[DeltaFactor(pins) → RAToTrainingFrame → FlowFactor → ProxyOffsetReparam → RAToEventFrame]`: single-pass sampling with preserved density (no Gibbs, no density recovery). The pinned values have one owner (the chain), and the heterodyne consumes the proxy through `prepared_data(conditioning=...)` — one method with a uniform row-aligned contract serving pins and sweeps alike.
- **Chirp-mass scan** (`--chirp-mass-scan`): sweeps the proxy over the training prior on the ordinary chain machinery (grid table root, per-row batched heterodyne, one network pass per block), selects the maximum-likelihood draw, and records the trigger + SNR in provenance. `--fixed-context-parameters` accepts the published demo-INI form.
- A pinned sky position is rotated into the network's training frame at input and back at output; exactly zero for networks whose reference time is the trigger (all current ones).

## Verification

- Legacy-vs-composed **bit-exactness** through both pipe stages (NPE and GNPE with density recovery), IS evidence identical to recorded digits; the legacy samplers are deleted (~1200 lines) with the last containing commit tagged `legacy-samplers-final`.
- BNS sampling **bit-exact to float32 against the `bns_add_dingo_pipe_max` research branch** on frozen event data; IS side verified live ≡ from-file bit-exactly.
- **Real GW170817** (cleaned LOSC frames + GWTC-1 PSDs) through the full pipe: with the demo pins, efficiency ~10–11%, log BF +510, chirp mass 1.19754 ± 0.00007; with *only* the sky pinned, the scan recovers the trigger (1.1976, SNR 33.0) and the evidence agrees with the pinned run within Monte Carlo error — the prior-conditioning invariance working in production.
- **GPU campaign (A100 server)**: full test suite green on Linux/CUDA; NPE sampling with the reverse-fold `log_prob` re-evaluated on-device (float32-level agreement); real GW170817 **BNS pin and scan runs reproduce the reference evidences within Monte Carlo error** (scan trigger 1.19756, snr 33.0, efficiency ~11%); GNPE with density-recovery training and `prior-dict-updates` exercised end-to-end on CUDA. Two device leaks in the previously CPU-only table-rooted chain path were found and fixed.
- A standing regression battery (full suite + pinned BNS/NPE/scan harnesses) ran after every change in this arc.

## Documentation

- **New pages**: [Sampling chains](https://github.com/dingo-gw/dingo/blob/sampler-revamp/docs/source/sampling_chains.md) -- the factorized-sampler concepts page (the product-of-conditionals model, step types, the sampler context, multiplicity rules, the reverse density fold, provenance, and chain building, with figures) -- and [Binary neutron stars](https://github.com/dingo-gw/dingo/blob/sampler-revamp/docs/source/bns.md) -- the DINGO-BNS workflow (phase heterodyning, prior conditioning, the chirp-mass scan, and the dingo_pipe configuration).
- **Rewritten**: [Inference](https://github.com/dingo-gw/dingo/blob/sampler-revamp/docs/source/inference.md) (the Python-API workflow page) and the [injection example](https://github.com/dingo-gw/dingo/blob/sampler-revamp/docs/source/example_injection.md), which was executed end to end against a trained GNPE pair as part of the update.
- **Updated**: [GNPE](https://github.com/dingo-gw/dingo/blob/sampler-revamp/docs/source/gnpe.md) (live inference routes replacing the removed `dingo_analyze_event`; a new single-step section) and [dingo_pipe](https://github.com/dingo-gw/dingo/blob/sampler-revamp/docs/source/dingo_pipe.md) (the BNS sampler options; the base-domain likelihood default).
- `build_model_from_kwargs` is re-exported from `dingo.core.posterior_models` so the documented import works. Build locally with `uv run --group docs sphinx-build -b html docs/source docs/build`.

## Review plan: importance sampling moves out of `Result` in a stacked PR

Following review feedback that `Result` still owns the importance-sampling orchestration, a **stacked follow-up PR** will externalize it: a domain-agnostic `ImportanceSampler` engine in `dingo/core`, a construction-only `GWImportanceSampler.from_result()` in `dingo/gw`, and the proposal-completion steps (synthetic phase, calibration) as ordinary chain factors. The evolving-prior state (`prior` / `geocent_time_prior` / `phase_prior` and the five order-coupled mutating methods) leaves `Result` entirely.

**Scheduled for deletion there — suggest reviewing these only lightly here** (they are behavior-preserving on this branch, covered by the A/B verification above; review effort pays off on their replacement instead): `Result.importance_sample`, `reset_event`, `_calculate_evidence`, `train_unconditional_flow` in `dingo/core/result.py`; `_build_likelihood`, `sample_synthetic_phase`, `sample_calibration_parameters`, `update_prior`, the marginalization-kwargs properties, and the `use_base_domain` setter in `dingo/gw/result.py`.

The dead `dingo/gw/importance_sampling/` package (an unregistered CLI plus a plotting module imported by nothing else) is deleted in this PR; the stacked PR repurposes that package name for the operator's GW assembly.

## Intended behavior changes

- Calibration spline nodes span each detector's event frequency range (as bilby_pipe places them) rather than the domain bounds, in `Result.sample_calibration_parameters`, `GWSignal`, and the two calibration transforms. Calibration results differ numerically from before for any run with a cropping or importance-sampling frequency range; sampling and likelihood agree with each other (`tests/gw/test_calibration_frequency_range.py`).
- Importance-sampling data are regenerated for a wider frequency range or another duration, on the network band extended to the request, and the likelihood works on that recorded grid. A multibanded model with a duration update now runs on the base domain (legacy raised `NotImplementedError`).
- `density-recovery-settings` rejects the legacy keys `remove_init_outliers`, `low_latency_label`, and `outdir` with an error.
- `dingo_pipe` gains `sampling-seed` (bilby_pipe's shape, plus torch); the sampling node records the seed it used under `settings["sampler"]`. `reset_event` keeps the pre-reset event record as `importance_sampling_metadata["proposal_event_metadata"]` and no longer writes the `updates` diff.

## Compatibility and deferred work

- Pre-v0.9.0 models need `compatibility/remove_domain_window_factor.py` (unchanged story).
- Deferred, tracked for follow-ups: Max's decimated/heterodyned fast likelihood (and the JAX layer from the paper), the synthetic-phase `compute_likelihood` fast path, embed-once on `FlowWrapper`, per-event time scans for pre-merger networks, CUDA/server realism runs.


stephengreen:
@max-dax — this is the factorized-sampler spine we've been coordinating on `sampler-revamp`, now at a good point to look at: Increments 1a–4 are done, `_post_process` is fully eliminated, and it's bit-exact vs the old samplers throughout. Opened as a draft for design feedback before any `dingo_pipe` wiring. Would value your eyes on the `Step` structure and the two `factors.py` modules whenever you get a chance — it supersedes the per-factor batching and folds in the GNPE kernel/flow split we worked through.

stephengreen:
@max-dax For reviewing, I would recommend starting by reading the [updated documentation](https://app.readthedocs.org/projects/dingo-gw/builds/33567755/), mainly the sections on "Sampling chains" and "Binary neutron stars".

dingo/gw/likelihood.py
```diff
@@ -512,7 +528,16 @@ def _log_likelihood_phase_marginalized(self, theta):
...
                 ]
             )
-            return self.log_Zn + ln_i0(np.abs(kappa2C)) - 1 / 2.0 * rho2opt
+            log_likelihood = self.log_Zn + ln_i0(np.abs(kappa2C)) - 1 / 2.0 * rho2opt
+
+            if self.return_aux_snr:
+                # Phase-maximized matched-filter SNR: |kappa2C| / sqrt(rho2opt).
+                # (ln_i0(|kappa2C|) is the phase-marginalized likelihood term
+                # above, not a matched-filter statistic.)
+                snr = np.abs(kappa2C) / rho2opt**0.5
```
stephengreen, line 537: @max-dax flagging this line for a look — it feeds the trigger SNR that the chirp-mass scan logs and stores in provenance.

The deep review caught that this previously reused the phase-marginalized likelihood term `ln_i0(|kappa2C|)` as the SNR numerator. The intended statistic is the phase-maximized matched-filter SNR: under the (2,2) approximation, `<d, h(phi)> = |kappa2C| cos(2 phi - arg kappa2C)`, so maximizing over phase gives `|kappa2C| / sqrt(rho2opt)` (with `rho2opt` phase-invariant). By contrast, `ln_i0(|kappa2C|) ~ |kappa2C| - 0.5 ln(2 pi |kappa2C|)` is the marginalized (integrated) weight — the maximized value minus the phase-volume Occam penalty — so the old value was biased low (GW170817 scan: 32.89 → 32.99 after the fix). Trigger *selection* was never affected, since the scan argmaxes the log-likelihood, which was always correct.

A regression test pins the statistic now (`test_phase_marginalized_aux_snr_is_phase_maximized`). Note `return_aux_snr` is new in this branch — it did not come from #354; the PR description's earlier attribution was wrong and has been corrected.

docs/source/sampling_chains.md
```diff
@@ -0,0 +1,361 @@
...
+that contains a `GibbsBlock` is therefore density-free (no log probability). For importance sampling, the density must be [recovered](result.md#density-recovery)
+after Gibbs sampling. Dingo uses this step only for multi-iteration [GNPE](gnpe.md).
+
+## Sampler context
+
+The `GWSamplerContext` holds event data and metadata, along with model metadata, and generates derived objects such as preprocessed data for networks and the GW likelihood. The context is shared across steps for consistency.
+
+### Outputs
+
+`prepared_data(conditioning=None)`
```
nihargupte-ph: 

docs/source/sampling_chains.md
```diff
@@ -0,0 +1,361 @@
...
+
+In practice, obtaining posterior samples is more complicated than just sampling from a flow. It is often also necessary, for instance, to add fixed parameters, apply coordinate transformations (sky rotations), iterate with GNPE, or reconstruct the synthetic phase (which requires access to the likelihood). The additional steps necessary for sampling in these cases are determined by model and event metadata. However, without a systematic organizing principle, the code to implement these steps can become unwieldy.
+
+Dingo's approach is the **factorized sampler**, which organizes sampling into an explicit chain of steps. The idea is to express the posterior as an ordered product of conditionals,
+
+$$
+q(\theta_1, \ldots, \theta_n | d) = \prod_i q_i(\theta_i | \theta_{<i}, d).
+$$
+
+A **factor** in this product represents a step, which could include, e.g., a normalizing flow, a Dirac delta function, or the phase posterior conditioned on the remaining parameters. In general a factor may be stochastic, such as a network, or it may be a point mass that pins parameters to fixed values. In addition to factors, there can also be **reparametrization** steps (e.g., the sky rotation).
```
nihargupte-ph, line 13: I think it could make sense to not introduce factors and reparametrizations here and insteadjust say that the object is that we stack Steps together. Later you could describe the types of Steps in the form of factors and repramaetrizations, however, in the first section differentiatiating between these types I don't think adds much.

dingo/core/inference/composer.py
```diff
@@ -0,0 +1,499 @@
...
+            if clobbered:
+                raise ValueError(
+                    f"A step producing {step.parameters} would overwrite existing "
+                    f"column(s) {sorted(clobbered)}. Only a Reparametrization may "
+                    f"replace columns (its inverse can rebuild them for log_prob)."
+                )
+            if isinstance(step, TargetCorrection):
+                # A correction has no inverse, so `log_prob` cannot rebuild what it
+                # consumes: it may consume side-channel intermediates only.
+                sampled_consumed = set(step.consumes) & sampled
```
nihargupte-ph, line 148: I felt the name step.consumes was a bit confusing since consumes implies that the column is both used and dropped. But really the variables in `self.consumes` are just dropped whereas the variables in `self.conditioning` denote what is being used. I think a better name could be `self.dropped` or `self.columns_dropped`.

dingo/gw/inference/steps.py
```diff
@@ -0,0 +1,665 @@
...
+        self.parameters = ["delta_log_prob_target"]
+        self.conditioning = list(kernel_factor.parameters) + list(
+            kernel_factor.gnpe_parameters
+        )
+        self.consumes = list(kernel_factor.gnpe_parameters)
+
+    def correction(self, given, context):
+        proxies = {p: given[p] for p in self.kernel_factor.parameters}
+        times = {k: given[k] for k in self.kernel_factor.gnpe_parameters}
+        correction = self.kernel_factor.log_prob(proxies, context, times)
```
nihargupte-ph, line 664:
```suggestion
        proxies = {p: given[p] for p in self.kernel_factor.parameters} # theta_hat
        gnpe_params = {k: given[k] for k in self.kernel_factor.gnpe_parameters} # theta
        correction = self.kernel_factor.log_prob(proxies, context, gnpe_params) # log p(theta_hat | theta)
```

nihargupte-ph (commented):
just adding a few comments for now will continue to read the PR

dingo/gw/result.py
```diff
@@ -255,7 +244,9 @@ def update_prior(self, prior_update):
             the Result and later instantiated.
         """
         self.importance_sampling_metadata["prior_update"] = prior_update.copy()
-        prior_update = PriorDict(prior_update)
+        # PriorDict instantiates in place, so work on a copy: the caller's dict
```
nihargupte-ph, line 247: Should the SamplerContext also be used to update the prior the same way the likelihood is?

dingo/gw/transforms/gnpe_transforms.py
```diff
@@ -227,3 +229,112 @@ def __call__(self, input_sample):
...
+            parameters = {**sample.get("parameters", {}), **extrinsic_parameters}
+            proxies = self.sample_proxies(parameters)
+            extrinsic_parameters.update(proxies)
+            delta_parameters = {
+                "delta_"
+                + p[: -len("_proxy")]: parameters[p[: -len("_proxy")]]
+                - proxies[p]
+                for p in self.proxy_list
+            }
+            extrinsic_parameters.update(delta_parameters)
```
nihargupte-ph, line 323: I'm not sure if it fits in this pr neccesarily, but we should eventually rename `extrinsic_parameters` to `augmentation_parameters` and `intrinsic_parameters` to `dataset_parameters`. Mainly because now the `delta_chirp_mass_proxy` is not really an `extrinsic_parameters` in the usual sense, but it is sampled on the fly like the rest of the extrinsic parameters

dingo/gw/inference/steps.py
```diff
@@ -0,0 +1,665 @@
...
+                y, log_prob = self.model.sample_and_log_prob(
+                    x["data"], x["context_parameters"], num_samples=num_samples
+                )
+            else:
+                y, log_prob = self.model.sample_and_log_prob(
+                    x["data"], num_samples=num_samples
+                )
+        # The network returns (n_rows, num_samples, dim): flatten with the draws for a
+        # row adjacent, and repeat the per-row extrinsic parameters (proxies, the
+        # preferred-proxy geocent time) to match before the post-network corrections.
```
nihargupte-ph, line 371:
```suggestion
        # The network returns (n_rows, num_samples, dim) where n_rows is the number of proxy parameters. However, we do not treat the proxy parameters differently in the subsequent steps. Therefore, we flatten the proxies such that each row of x["parameters"] is a posterior sample. 
```

dingo/core/inference/composer.py
```diff
@@ -0,0 +1,525 @@
...
+            per drawing step, in chain order, so that a later step may draw several
+            samples for each row it receives (for example several extrinsic draws
+            per intrinsic sample). Steps that do not draw (pins, tables,
+            reparametrizations, target corrections) take no count. A chain with no
+            drawing step at all (a stored table run through a reparametrization,
+            say) runs once, with `num_samples=1`.
+        context : SamplerContext
+            The per-event shared state.
+        batch_size : int, optional
+            Chunk the first count into batches of this size, to cap the peak memory.
```
nihargupte-ph, line 189: I am wondering if TargetCorrection should be excluded here. Because TargetCorrection emits `delta_log_prob_target` as it's parameter, but that is a side channel. So like in principle we can compute the log prob of a sample which doesn't have `delta_log_prob_target` which is why it shouldn't be a sampled parameter (I think).

nihargupte-ph (commented):
Added a few more comments.

nihargupte-ph:
@stephengreen — a thought on metadata for the stacked `ImportanceSampler` PR.

Right now, when one does `reset_event` the `event_metadata` used for the sampling job gets overwritten by the IS `event_metadata`. It can be partially recovered from `importance_sampling_metadata["updates"]`, but only as a diff. On this branch that diff then drives `GWSamplerContext.derive()`, which rebuilds the data domain from the update keys — and that path can't actually change the frequency range (the `domain_dict` intersection at `context.py:341-344` never picks up `minimum_frequency`/`maximum_frequency`, and multibanded domains raise). We hit this in ProdPE: an `importance sampling updates: {maximum frequency: ...}` request in the asimov blueprint was silently ignored, so IS ran at the network's f_max and the calibration spline nodes ended up at different frequencies from Bilby's.

Maybe a cleaner option is for the `Result` to carry two metadatas, `proposal_metadata` and `target_metadata`. Everything in `Result` would be initialised from `target_metadata` only: we instantiate p(d|θ) p(θ) from it, so importance sampling to a different likelihood is clean. `target_metadata` would start as a subset of `proposal_metadata` and then be updated with the `importance_sampling_updates`. The likelihood domain would come from the IS event dataset itself (array length, `T`, per-detector frequency range in its settings) rather than from update keys, so f_min/f_max, duration and PSD all follow the same path.

dingo/gw/transforms/gnpe_transforms.py
stephengreen: I agree that all the keys are a bit off by now (especially `extrinsic_parameters`, but also `waveform` meaning "data"). I suggest to raise an Issue and we can fix this in the future.

dingo/gw/result.py
stephengreen: Deliberately not, at least not here. The context owns what the trained model and the event fix: the static training prior, the data representation, and the likelihood machinery. Choices about the target density are applied downstream: the marginalization settings go in as arguments to `likelihood(...)`, and the prior update and the time/phase split-off are applied by `Result`. Two things keep the prior out of the context. The split-off depends on which columns the samples carry, which the context never sees. And the chain itself needs the training prior (the delta-prior fillers, the chirp-mass scan), not an updated one.

`Result` holding a prior at all is transitional: right now `Result` *is* the importance-sampling operator. In the stacked IS PR the assembly builds the target prior from the context's static prior plus the update and restriction, hands it to the operator alongside the likelihood, and `Result` becomes passive. So this will be handled by the Result refactor rather than by the context.

docs/source/sampling_chains.md
stephengreen: Agreed. In 9e01bf3d the introduction describes the chain in terms of steps only (each conditional in the product is a step, and a step may also be a change of variables) and leaves factors, reparametrizations, and corrections to the step-types section.

dingo/core/inference/composer.py
stephengreen: You were right that the name was hiding something. `consumes` was doing two unrelated jobs: for a reparametrization it marked the inputs the bijection replaces (expressed as conditioning minus parameters, with overrides), and for the kernel correction it was cleanup of the recomputed detector times.

In 9e01bf3d it is gone. A `Reparametrization` now declares `inputs` (the columns it transforms, replaced by its outputs) and `conditioning` (read-only, stays in the chain); the composer reads both, drops the inputs after the step, and `inverse` receives only the conditioning. Nothing else removes columns: side channels, i.e. columns a factor emits beyond its `parameters` such as the detector times, are dropped by the composer at the end of the chain.

dingo/core/inference/composer.py
stephengreen: Yes, that was an inconsistency (I take this to be about the `sampled` set in chain validation, which was at line 158). The correction listed `delta_log_prob_target` under `parameters`, so the set picked it up as a sampled column. In 9e01bf3d a `TargetCorrection` has empty `parameters` and declares its annotation column in `produces`. The `sampled` bookkeeping is gone altogether, along with the rule that a correction may consume only side channels, since corrections no longer remove anything (see the `consumes` thread).

dingo/gw/inference/steps.py
stephengreen: Applied in 9e01bf3d (black-formatted).

dingo/gw/inference/steps.py
stephengreen: Reworded in 9e01bf3d. One correction to the suggestion: `n_rows` is the number of conditioning rows (one per proxy sample), not the number of proxy parameters. I kept the note about repeating the per-row extrinsic parameters, which the post-network transforms rely on.

stephengreen:
Thanks, this led somewhere useful. The diagnosis was slightly off, though: the domain-dict intersection in `derive` is the legacy `_rebuild_domain` ported verbatim, and for a uniform domain it only ever changed `delta_f`. Frequency ranges have been applied by ASD masking from the event metadata since #287. What actually dropped your request was the pipe: `data_generation.py` popped the frequency keys from `importance_sampling_updates` before generating the IS data, so the IS event settings carried the network's range, `reset_event` saw no change, and the likelihood masked at the network's band. Identical on `main`. A second layer explains the calibration nodes: they were placed over the domain bounds in both calibration paths, whereas bilby_pipe places them over each interferometer's analysis range.

Both are fixed in 06a9d746, following the rule you suggested: the likelihood is derived from the event dataset, not the network. Data generation now records the grid it generated the data on (the network's band extended to the requested range, at the requested duration), the likelihood builds its grid from that record and extends the waveform generator to contain it, and the calibration nodes span each detector's recorded range. `derive` and the context's representation state are gone; `use_base_domain` is an argument of `likelihood()`. Widening beyond a `domain_update` band works; I left the same explanation on #406 with a mapping to `lvk-stable`.

On the two metadata blocks: I agree with the direction but would rather not name the primary block "proposal" when no importance sampling happens. Keeping `event_metadata` as the event analyzed and snapshotting the sampling-time record under the importance-sampling block gives the same information. I'll take that up with the Result rework.

stephengreen:
@nihargupte-ph On the two metadata blocks: fb493b8e takes the direction you proposed, in the shape described above. `reset_event` now keeps the full pre-reset event record as `importance_sampling_metadata["proposal_event_metadata"]` (the first record wins on repeated calls) and no longer writes the `updates` diff, which nothing has read since the frequency-range change. `event_metadata` remains the event analyzed, so PESummary and asimov see no change. The proposal's event data are not kept with it; they stay in the sampling-stage file, which is enough for now. The deeper reorganization (the sampler block, the synthetic phase's absence from the chain listing, the mix of settings in the importance-sampling block) waits for the Result rework in the stacked PR.

stephengreen:
Also a9e0c450: `dingo_pipe` gains a `sampling-seed` option, since the sampling and importance-sampling jobs seeded nothing. Both nodes carry a `sampling_seed` property mirroring bilby_pipe's `DataAnalysisInput`, plus torch. As in bilby_pipe, importance-sampling job n uses the seed plus n, and each job draws and logs its own when the ini has none. The sampling node records the seed it used under `settings["sampler"]`; the importance-sampling block records nothing, so parallel parts still merge. With a seed the sampling node is bit-reproducible, and importance sampling is with a single process only: the Pool workers are not re-seeded, which predates this branch and is written up in #408.

stephengreen:
@AlexandreGoettel flagging an asimov decision made in 5e94ed19, in case you want it the other way.

The asimov INI-consistency test (`tests/asimov/test_parser_ini_consistency.py`) was failing on this branch: two new `dingo_pipe` options, `chirp-mass-scan` and `fixed-context-parameters` (single-step GNPE / DINGO-BNS: pin the model's context parameters, or determine the chirp mass from the data before sampling), were in neither `dingo/asimov/dingo.ini` nor `EXCLUDED_FROM_INI`. This had gone unnoticed because the pytest workflow only runs against `main`.

They are now in the exclusion set, i.e. not exposed through the asimov template, since no BNS network is distributed yet. If asimov should be able to drive BNS runs, they would instead go into the template with a conditional block like the one for `importance-sampling-updates`. Shout if you'd rather have that now.

nihargupte-ph:
I think it's already been stated in the PR somewhere, but sample calibration paramteres should probably be part of the sampling chain. It can be a seperate PR though, we should not try to make this one any longer! And actually it should come before the synthetic phase. I say this because if we want to use the likelihoods from the synthetic phase for the IS, those posterior samples need to include the calibration parameters.

nihargupte-ph (approved):

stephengreen:
> I think it's already been stated in the PR somewhere, but sample calibration paramteres should probably be part of the sampling chain. It can be a seperate PR though, we should not try to make this one any longer! And actually it should come before the synthetic phase. I say this because if we want to use the likelihoods from the synthetic phase for the IS, those posterior samples need to include the calibration parameters.

Thanks, I've created #413 to remember this.

PR #393, opened 2026-07-03
feat(gw): time-domain whitened-strain PPD plots on Result

nihargupte-ph:
## Summary

Adds time-domain whitened-strain **posterior predictive distribution (PPD)** plotting to `dingo.gw.result.Result`.

- New `Result` methods to generate strain PPD plots (not a standalone module)
- Min/max envelope with nested credible-level bands, plus overlay of the max-probability waveform
- Per-(mode, detector) stacked min/max panels
- Plotting helpers in `dingo/gw/utils/plotting.py`; injection support in `dingo/gw/injection.py`
- Design spec under `docs/specs/2026-06-30-strain-ppd-plotting-design.md`
- Test at `tests/gw/test_ppd.py`
- `.gitignore`s a `local/` scratch dir for generated images

## Test plan

- [x] `tests/gw/test_ppd.py`
- Iterated visually against the GW150914 dingo-ci result fixture


docs/superpowers/specs/2026-06-30-strain-ppd-plotting-design.md
lorenzopompili00, line 1: Should this file be removed?

dingo/gw/utils/plotting.py
```diff
@@ -0,0 +1,152 @@
...
+        d_t, d_td = data_td[ifo]
+        ax.plot(d_t, d_td, color=_DATA_COLOR, lw=0.8, alpha=0.6, zorder=1)
+        for i, (lvl, k) in enumerate(ordered):
+            if k < 1:
+                continue
+            band = td[:k]
+            ax.fill_between(
+                tt, band.min(axis=0), band.max(axis=0),
+                color=color, alpha=0.25 + 0.22 * i, zorder=2 + i,
+                label=(f"{lvl:.0%} CI" if (multi and row == 0 and lvl is not None) else None),
```
lorenzopompili00: Just noting that this is a different statistic than what bilby (and most LVK reconstruction plots?) show, so I would avoid labelling it, e.g., "90% CI". The method `plot_interferometer_waveform_posterior` in bilby plots the pointwise 90% band: at each time sample, the central 90% of posterior waveforms. Here we plot the min/max envelope of draws from the 90% HPD set in parameter space, which is a different quantity.  Both are reasonable things to plot, but with the same "90% CI" label someone comparing against a bilby plot of the same event might conclude the posteriors disagree when it's only the band definition. I would suggest labelling it "90% HPD" or "90% HPD envelope" (or similar), and maybe mentioning the difference with Bilby in the docstring.

dingo/gw/utils/plotting.py
```diff
@@ -0,0 +1,152 @@
...
+    zoom = zoom if zoom is not None else (-1.0, 0.2)
+
+    t0 = 1 / (2 * domain.delta_f)  # merger at segment midpoint -> t = 0
+    phase_shift = np.exp(2j * np.pi * domain() * t0)
+
+    # Grey whitened-data trace per detector (shared across that detector's panels).
+    data_td = {}
+    for ifo in ifos:
+        d_times, d_x = one_sided_fd_to_td(data_fd[ifo] * phase_shift, domain)
+        data_td[ifo] = (d_times - t0, np.convolve(np.real(d_x), np.ones(4) / 4, mode="same"))
```
lorenzopompili00: Is this needed? Why not just plot the raw time domain data (as done by Bilby)?

dingo/gw/utils/plotting.py
```diff
@@ -0,0 +1,152 @@
...
+        if multi and row == 0:
+            ax.legend(loc="upper right", fontsize=8, framealpha=0.9)
+
+    axes[-1].set_xlabel("time to merger (s)")
+    fig.savefig(filename, dpi=200, bbox_inches="tight")
+    plt.close(fig)
+    return axes
+
+
+def one_sided_fd_to_td(
```
lorenzopompili00: The docstring says the sqrt N normalization makes whitened noise have unit variance, but it comes out at sqrt 2, while and bilby's convention is exactly 1, so our y-axis is sqrt 2 larger than a bilby "whitened strain" plot of the same data. Since data and model go through the same function the plot shape is fine,  it's just a scale issue.

A quick check from Claude:
```python
import numpy as np
rng = np.random.default_rng(0)
n = 4097  # one-sided bins, f_max = (n-1)*df
fd = rng.normal(size=(500, n)) + 1j * rng.normal(size=(500, n))

# DINGO one_sided_fd_to_td
fd_d = fd.copy(); fd_d[:, 0] = 0
n_time = 2 * n - 1
td_dingo = np.fft.irfft(fd_d, n=n_time, axis=-1) * np.sqrt(n_time)

# Bilby get_whitened_time_series_from_whitened_frequency_series (full mask)
mask = np.ones(n, dtype=bool); mask[0] = False
fd_b = fd.copy(); fd_b[:, ~mask] = 0
td_bilby = np.fft.irfft(fd_b, axis=-1) * mask.sum()**0.5 / mask.mean()

print(td_dingo.std())                 # 1.414
print(td_bilby.std())                 # 1.000
print(td_dingo.std() / td_bilby.std())  # ≈ sqrt(2)
```

More broadly, this makes me think if one could reuse more of the bilby functionality instead of reimplementing things. For example, there is a method [get_whitened_time_series_from_whitened_frequency_series](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/detector/interferometer.py#L702) which already does what we need, and we already use the bilby Interferometer class so this wouldn't add extra dependencies.

dingo/gw/injection.py
```diff
@@ -442,3 +444,19 @@ def random_injection(self):
...
         }  # Some parameters are np.float64
         return self.injection(theta)
+
+
+def safe_signal(gw_signal: "GWSignal", theta: dict) -> Optional[dict]:
+    """Call ``gw_signal.signal(theta)``, returning ``None`` if generation fails.
+
+    Even in-prior posterior draws can hit waveform-model edge cases -- e.g. SEOBNRv5EHM's
+    eccentric interpolation failing on an interior parameter, or simply a waveform backend
+    (pyseobnr / lalsimulation) that differs from the one used when the result was produced.
```
lorenzopompili00: This docstring feels a bit too specific? I'm not sure there's any realistic case where using a different version of pyseobnr / lalsimulation would have waveform failing that used to generate tbh

dingo/gw/result.py
```diff
@@ -781,3 +785,260 @@ def pesummary_prior(self):
...
+        filename: str = "ppd_td.png",
+        credible_interval: Union[float, Sequence[float]] = 0.9,
+        num_waveforms: int = 1000,
+        num_processes: int = 1,
+        zoom: Optional[Tuple[float, float]] = None,
+    ) -> Tuple[dict, dict, dict]:
+        """Plot the time-domain whitened-strain posterior-predictive distribution.
+
+        For each mode and detector, inverse-FFTs whitened waveforms to the time domain with
+        the merger at t = 0 and shades the pointwise min/max envelope, with the
```
lorenzopompili00: Merger at t=0 reads very generic: what does "merger" mean exactly? Is it the trigger time, or the time at which the waveform peaks? Is it defined at the geocenter, or at the detector? The docstring (and label on the plot) could be more specific.

dingo/gw/result.py
```diff
@@ -781,3 +785,260 @@ def pesummary_prior(self):
...
+        Ranking by the density and accumulating ``m_i`` until the cumulative mass reaches X
+        keeps exactly the samples above the density threshold ``c``
+
+        Prior filter. Samples are first restricted to the prior support (finite ``log_prior``).
+        The flow proposal is a smooth density over a box and leaks a small fraction of draws
+        outside the prior (e.g. tiny negative spin magnitudes); these are unphysical, carry
+        zero target mass, and account for most waveform-generation failures. This is exactly
+        the in-prior restriction that :meth:`importance_sample` applies.
+
+        Per-sample guard. The prior filter is not sufficient on its own: a waveform model can
```
lorenzopompili00: Again, the comment seems overly specific

dingo/gw/result.py
```diff
@@ -781,3 +785,260 @@ def pesummary_prior(self):
...
+                credible_idx = rng.choice(
+                    credible_idx, size=num_waveforms, replace=False
+                )
+            # Order the draws by descending density so the plot can take the densest top-k as
+            # the inner (tighter) credible level.
+            credible_idx = credible_idx[np.argsort(rank[credible_idx])[::-1]]
+            sel_rank = rank[credible_idx]
+            theta = self.samples.iloc[credible_idx]
+
+            # The prior filter removes out-of-prior draws, but it does not guarantee every
```
lorenzopompili00: Again, the comment seems overly specific

dingo/gw/utils/__init__.py
lorenzopompili00, line 1: Would it make sense to combine this with `dingo.gw.gwutils`?

lorenzopompili00 (changes requested):
Thanks @nihargupte-ph! I left a few small comments.

Also, would it be possible to have an example of the figure that gets generated? It might also be nice to have this incorporated in the documentation.

docs/superpowers/specs/2026-06-30-strain-ppd-plotting-design.md
nihargupte-ph: yes it should be removed

nihargupte-ph (commented):
Thanks for the review! I made a few changes.

nihargupte-ph:
Here's a few images: 
ppd_GW200129_bands_draws

ppd_S230630bq_bands_fullsegment
ppd_S230630bq_bands_draws
ppd_GW200129_bands_fullsegment

dingo/gw/utils/plotting.py
nihargupte-ph: True, I've now edited it to follow the LVK convention to avoid confusion. In particular for each t we draw the 90% interval essentially using a histogram. The one issue with this is if there is a multimodal p(h(t) | d). I did play around with drawing the contours multimodally and also trying to plot the density as a heatmap but it was not coming out good visually

dingo/gw/utils/plotting.py
nihargupte-ph: For longer signals, the convolution is basically a slighting window average which lets you visual the strain data better (it's like supressing the noise). Although, maybe this should be a feature that users can implement themselves and by default we do not convolve.

dingo/gw/utils/plotting.py
nihargupte-ph: True, the updated PR fixes this using bilby functions whenever possible.

dingo/gw/injection.py
nihargupte-ph: Agreed, I've edited it now to make it more general.

dingo/gw/result.py
nihargupte-ph: Maybe a way to be agnostic about this is to report the GPS time but not refer to a specific merger time.

dingo/gw/result.py
nihargupte-ph: Agreed.

dingo/gw/result.py
nihargupte-ph: agreed.

dingo/gw/utils/__init__.py
nihargupte-ph: Hmmm maybe, I think eventually we could use more plotting utilities so maybe makes sense to have a new file.

PR #396, opened 2026-07-08
ENH: Add get_one_dimensional_median_and_error_bar in result

filippo-santoliquido:
I've added a new method to [`dingo/core/result.py`](https://github.com/dingo-gw/dingo/compare/main...filippo-santoliquido:dingo-ET:feature/result-percentile?expand=1#diff-f2e161644a8b69e9d915ad15f9552062d5a98cd1fe16013d29ef0e6df7464a2e) that computes the one-dimensional median and error bar. It mirrors the equivalent method in [`bilby`](https://github.com/bilby-dev/bilby/blob/main/bilby/core/result.py#L1014-L1049), with the addition of support for importance weights.

### DIFF
--- a/dingo/core/result.py
+++ b/dingo/core/result.py
@@ -923,6 +923,84 @@ class Result(DingoDataset):
         else:
             return np.nan
 
+    def get_one_dimensional_median_and_error_bar(self, key: str, weighted: bool = False, fmt: str = '.2f', quantiles: list = [0.16, 0.84]):
+        """ 
+        Calculate the median and error bar for a given key
+
+        Parameters
+        ==========
+        key: str
+            The parameter key for which to calculate the median and error bar
+        weighted: bool, optional
+            Whether to use sample weights in calculating the median and error bar
+        fmt: str, ('.2f')
+            A format string
+        quantiles: list, 
+            A length-2 list of the lower and upper-quantiles to calculate the errors bars for.
+
+        Returns
+        =======
+        summary: namedtuple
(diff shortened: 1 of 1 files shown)


cmfabbri (changes requested):
Hi @filippo-santoliquido!

In `get_one_dimensional_median_and_error_bar`, `summary` is created via `namedtuple` in line 947, but this it's never actually instantiated. The code then does `summary.median = ..., summary.plus = ..., summary.minus = ...`, which sets attributes directly on the class object. Also notice that `minus` and `plus` are used here, while `lower` and `upper` where used when you created `summary`.
The file already has the correct pattern for this a bit further down (`Pvals = namedtuple(...) / pvals = Pvals(...)` around line 1115-1120 in this branch), could you follow that instead? It would be something like:
`summary_tuple = namedtuple('summary', ['median', 'lower', 'upper', 'string'])
...
return summary_tuple(median=quants[1], lower=quants[2]-quants[1], upper=quants[1]-quants[0], string=string)`,
so no need to add the attributes as you were doing in lines 961-963.

filippo-santoliquido:
Thanks a lot @cmfabbri ! I've pushed a commit addressing your comments:
- `summary` is now instantiated properly and returned as a namedtuple instance.
- Field names are now consistent throughout (`median`, `lower`, `upper`, `string`)
- While I was in there I also tidied up the docstring. 

Let me know if anything else needs changing.

cmfabbri (approved):

PR #402, opened 2026-08-25
Add Dingo-T1 transformer-based embedding network

annalena-k:
## Summary

This PR introduces **Dingo-T1**, a transformer-based architecture for gravitational-wave parameter estimation. A single trained model can adapt to a wide range of analysis configurations at inference time. The code corresponds to results reported in the paper [Kofler et al. 2026, PRL 136, 251401](https://link.aps.org/doi/10.1103/n3ld-b4wv), only small implementation details are changed.

### New architecture

- **`StrainTokenization`** (`tokenization_transforms.py`): converts per-detector strain into a flat sequence of fixed-size frequency-domain tokens, each carrying its data, position `(f_min, f_max, detector)`, and a mask bit.
- **`DenseResidualNet`** (`resnet.py`): a new ResNet variant with optional `layer_norm`, used as the NSF coupling conditioner when `layer_norm=True`. The original glasflow `ResidualNet` is kept as default for checkpoint compatibility (the two are not interchangeable: GLU-per-block vs. one-time context concatenation).
- **`TransformerModel`** (`core/nn/transformer.py`): `Tokenizer` → `TransformerEncoder` → pooling → `DenseResidualNet` flow head. Wired into `normalizing_flow`, `cflow_base`, `build_model`, and `train_pipeline` via `embedding_type: transformer`.

### Training-time augmentations

Four new transforms teach the model to handle missing or restricted data:

| Transform | What it trains | Settings key |
|-----------|---------------|--------------|
| `MaskFrequencyEdges` | Variable f_min / f_max per detector | `mask_frequency_edges` |
| `MaskFrequencyInterval` | Interior frequency notches | `mask_frequency_interval` |
| `MaskDetectors` | Absent detectors | `mask_detectors` |
| `RandomTokenMask` | Random token dropout | `mask_random` |

All four are optional, independent, and wired into `set_train_transforms`.

### Inference-time flexibility (dingo_pipe)

- **Per-detector frequency range**: `minimum-frequency` and `maximum-frequency` flags in the `.ini` file accept a single float or a per-detector dict; `GWSampler.minimum_frequency` / `maximum_frequency` properties rebuild the transform chain on assignment. Out-of-distribution warnings are raised when the requested range falls outside the training envelope.
- **PSD notching**: `--psd-notch-dict` ini flag sets ASD = `HIGH_ASD_VALUE = 1` in specified bins during data generation. At sampling time, notched regions are auto-detected from the stored ASD (`detect_asd_notches`) and the overlapping tokens are masked. An Asimov path (pre-notched ASD, no ini flag needed) is also supported. Importance sampling requires no special handling since ASD = 1 ≫ real noise.
- **Detector updates**: `dingo_pipe` updates the active detector set from the event data at inference time.

### Hardware fix

`torch.backends.mha.set_fastpath_enabled(False)` is called in `TransformerModel.__init__` to avoid a crash with B200 CUDA kernels. The fast path does not accelerate our use case (we pass `src_key_padding_mask`), so there is no performance cost.

### Documentation

New `docs/source/example_dingo_t1.md` covers training and all inference-time options. Updated `dingo_pipe.md` and `inference.md`. Added `Kofler:2026fgw` to `refs.bib` and cited it in `index.md`.

## Known limitations / open questions

- **GNPE + transformer**: GNPE is currently only compatible with `embedding_type: resnet`. The `autocomplete_model_kwargs` function infers `gnpe_proxy_dim` from `data_sample[2]`, which for the transformer path is the position tensor, not GNPE proxies. Transformer + GNPE is deferred as future work.
- **Naming**: local variables `blocks` → `detector_indices` (and `num_blocks` → `num_detectors`, `num_tokens_per_block` → `num_tokens_per_detector`) in all `__call__` methods of `tokenization_transforms.py`. This is more explicit for the GW use case but may be too domain-specific for `dingo.core` — open to discussion.
- **AMP**: automatic mixed precision is not yet implemented (not specific to the transformer path; tracked separately).


dingo/core/nn/nsf.py
```diff
@@ -103,9 +114,28 @@ def create_base_transform(
...
+                hidden_dims=_hidden_dims,
+                activation=activation_fn,
+                dropout=dropout_probability,
+                batch_norm=batch_norm,
+                layer_norm=True,
+                context_features=_ctx,
+            )
+        else:
+            # glasflow's ResidualNet is kept here for backward compatibility: it
+            # concatenates context with the input rather than using GLU gating, so the
```
stephengreen: "rather than using GLU gating" isn't quite correct: the `ResidualBlock` in `nflows` still has the GLU gating. It has both types of conditioning. Also the various layers are different, so the classes aren't fully compatible.

dingo/core/nn/nsf.py
```diff
@@ -103,9 +114,28 @@ def create_base_transform(
...
             )
-        return transforms.PiecewiseRationalQuadraticCouplingTransform(
-            mask=mask,
-            transform_net_create_fn=(
+
+        if layer_norm:
+            # DenseResidualNet supports layer_norm and injects context via GLU gating
+            # in each residual block, rather than concatenating it to the input once.
+            _hidden_dims = tuple([hidden_dim] * num_transform_blocks)
+            _ctx = context_dim
```
stephengreen: Inline `_hidden_dims` and *especially* `_ctx`.

dingo/core/nn/resnet.py
```diff
@@ -0,0 +1,209 @@
...
+            whether to use batch normalization
+        layer_norm : bool
+            whether to use layer normalization
+        """
+        super().__init__()
+        self.input_dim = input_dim
+        self.output_dim = output_dim
+        self.hidden_dims = hidden_dims
+        self.num_res_blocks = len(self.hidden_dims)
+
```
stephengreen, line 176: Restore these lines from `dingo-t1`. They're needed to get proper normalization in the coupling flow and match `nflows`. Without this line, the `dingo-t1` demo network is incompatible.
```suggestion
        # This attribute is required by nflows.
        if all([d == self.hidden_dims[0] for d in self.hidden_dims]):
            self.hidden_features = self.hidden_dims[0]
```
With this line, we should also get better initialization for training new networks. Note that any networks trained on the `transformer` branch will need to be retrained.

dingo/core/nn/nsf.py
```diff
@@ -103,9 +114,28 @@ def create_base_transform(
             mask = nflows.utils.create_alternating_binary_mask(
                 param_dim, even=(i % 2 == 0)
             )
-        return transforms.PiecewiseRationalQuadraticCouplingTransform(
-            mask=mask,
-            transform_net_create_fn=(
+
+        if layer_norm:
+            # DenseResidualNet supports layer_norm and injects context via GLU gating
+            # in each residual block, rather than concatenating it to the input once.
```
stephengreen: "rather than concatenating it to the input once" -- `glasflow` actually does both

dingo/pipe/main.py
```diff
@@ -165,6 +179,16 @@ def fill_in_arguments_from_model(args, perform_arg_checks=True):
         frequency_input.minimum_frequency_dict,
         frequency_input.maximum_frequency_dict,
     )
+    if flexible_detectors:
+        check_detector_update(model_metadata=model_metadata, detectors=detectors)
+    else:
```
stephengreen: no-op. In the case of no `flexible_detectors`, we've already set `detectors` based on the `data_settings`.

dingo/pipe/main.py
```diff
@@ -90,7 +102,7 @@ def fill_in_arguments_from_model(args, perform_arg_checks=True):
         "duration": domain.duration,
         "minimum_frequency": domain.f_min,
         "maximum_frequency": domain.f_max,
-        "detectors": data_settings["detectors"],
+        "detectors": detectors,
```
stephengreen: `model_args` should describe the args from the model, so don't replace with the user-specified args.

dingo/pipe/main.py
```diff
@@ -80,6 +80,18 @@ def fill_in_arguments_from_model(args, perform_arg_checks=True):
...
+    # from the command-line args so the rest of the pipeline uses the right detectors.
+    detectors = data_settings["detectors"]
+    tok = data_settings.get("tokenization", {})
+    flexible_detectors = "mask_detectors" in tok or "mask_random_tokens" in tok
+    if flexible_detectors:
+        if args.detectors is not None:
+            detectors = [d.strip("'") for d in args.detectors]
+        elif args.channel_dict is not None:
+            detectors = list(convert_string_to_dict(args.channel_dict).keys())
+
```
stephengreen: Use `check_detector_update()` instead.

dingo/pipe/main.py
```diff
@@ -156,7 +168,9 @@ def fill_in_arguments_from_model(args, perform_arg_checks=True):
         setattr(args, k, v)
 
     frequency_input = Input([], [], print_msg=False)
-    frequency_input.detectors = model_args["detectors"]
+    # Frequency validation always uses the model's full detector list; the flexible
+    # detector subset only affects which data blocks are loaded at inference time.
+    frequency_input.detectors = data_settings["detectors"]
```
stephengreen: Would be better to only have to check against detectors we are planning to use.

dingo/pipe/main.py
```diff
@@ -192,7 +216,11 @@ def fill_in_arguments_from_model(args, perform_arg_checks=True):
         importance_sampling_updates = {
             k.replace("-", "_"): v for k, v in importance_sampling_updates.items()
         }
-    return {**changed_args, **importance_sampling_updates}, model_args
+    return (
+        {**changed_args, **importance_sampling_updates},
+        model_args,
+        flexible_detectors,
```
stephengreen: `flexible_detectors` never used. Remove.

dingo/gw/inference/gw_samplers.py
```diff
@@ -65,21 +67,58 @@ def __init__(self: SamplerProtocol, **kwargs):
...
     def detectors(self: SamplerProtocol):
-        return self.base_model_metadata["train_settings"]["data"]["detectors"]
+        if self._detectors is None:
+            self._detectors = self.base_model_metadata["train_settings"]["data"][
+                "detectors"
+            ]
+        return self._detectors
+
+    @detectors.setter
+    def detectors(self: _GWMixinProtocol, value: list[str]):
```
stephengreen, line 89: Validate the detector set here with `check_detector_updates()`, paralleling the frequency updates.

dingo/pipe/main.py
stephengreen: Also, allow the user to specify partial frequency restrictions, defaulting to model settings for unspecified detectors in the dict.

dingo/gw/inference/gw_samplers.py
```diff
@@ -671,12 +858,151 @@ def check_frequency_updates(
...
+    p_mask_hlv = mask_detector_settings["p_mask_hlv"]
+    for det in detectors_event:
+        if det not in p_mask_hlv:
+            raise ValueError(f"Detector {det} not included in p_mask_hlv={p_mask_hlv}.")
+        if p_mask_hlv[det] == 0.0:
+            raise ValueError(
+                f"Probability of keeping detector {det} is 0 in "
+                f"p_mask_hlv={p_mask_hlv}."
+            )
+
```
stephengreen: This has inverted logic. `p_mask_hlv` is the probability of masking, not keeping.

dingo/gw/inference/gw_samplers.py
```diff
@@ -671,12 +858,151 @@ def check_frequency_updates(
...
+        tok = model_metadata["train_settings"]["data"]["tokenization"]
+        if "mask_detectors" in tok:
+            _validate_detectors_transformer(
+                detectors_event=detectors,
+                detectors_network=detectors_network,
+                mask_detector_settings=tok["mask_detectors"],
+            )
+        elif "mask_random_tokens" in tok:
+            # Token-level masking does not constrain which detectors are present.
+            pass
```
stephengreen: If neither of these masks are specified, the check is passed (a silent failure).

dingo/gw/inference/gw_samplers.py
stephengreen: Also, fix associated tests.

dingo/core/nn/resnet.py
```diff
@@ -0,0 +1,209 @@
...
+    used to embed high dimensional input to a compressed output. Linear
+    resizing layers are used for resizing the input and output to match the
+    first and last hidden dimension, respectively.
+
+    Compared to glasflow.nflows.nn.nets.ResidualNet, which the residual blocks here are
+    based on, this implementation differs in two important ways:
+
+    1. **Context conditioning**: glasflow's ResidualNet concatenates the context vector
+       with the input once at the start of the network. Here, context is injected into
+       every residual block via a gated linear unit (GLU), giving the network repeated
```
stephengreen: Slight mis-statement. The `ResidualNet` also does the GLU (it does both).

dingo/gw/inference/gw_samplers.py
```diff
@@ -153,6 +262,10 @@ def event_metadata(self):
     def event_metadata(self, value):
         if value is not None:
             value = value.copy()
+            # Process detectors first so that frequency validation (which uses
+            # self.detectors) already reflects the event's detector subset.
+            if "detectors" in value and value["detectors"] is not None:
+                self.detectors = value.pop("detectors")
```
stephengreen: Popping `detectors` here (like the frequencies), but they should also be put back in `event_metadata` by the getter. This maintains provenance. The popping itself allows the objects to be mutated.

dingo/gw/inference/gw_samplers.py
```diff
@@ -671,12 +858,151 @@ def check_frequency_updates(
...
+            )
+        if crop_settings is not None:
+            _validate_maximum_frequency(f_max, detectors, domain, crop_settings)
+        if mask_frequency_edges_settings is not None:
+            _validate_maximum_frequency_transformer(
+                f_max, detectors, domain, mask_frequency_edges_settings
+            )
+
+
+def _validate_detectors_transformer(
```
stephengreen, line 837: Can this function be condensed? It seems long for what it does.

stephengreen:
I've reviewed part of this PR so far, primarily focusing on how detector updates are checked / propagated. Also I found a couple issues with the `DenseResidualNet` implementation. 

I can go ahead and implement these changes @annalena-k but let me know if I misunderstood something.

I also added backwards compatibility to be able to run `dingo-t1` trained networks (e.g., on Zenodo and the demo).

dingo/gw/inference/gw_samplers.py
```diff
@@ -671,12 +858,151 @@ def check_frequency_updates(
...
+    )
     detectors = model_metadata["train_settings"]["data"]["detectors"]
     domain = build_domain_from_model_metadata(model_metadata, base=True)
     if not isinstance(domain, (UniformFrequencyDomain, MultibandedFrequencyDomain)):
         raise ValueError("Frequency updates only possible for frequency domains.")
 
     if f_min is not None:
-        _validate_minimum_frequency(f_min, detectors, domain, crop_settings)
+        if not has_flexible_freq:
+            raise ValueError(
```
stephengreen: This will always fire for non-flexible models, even when the frequency is unchanged

dingo/gw/inference/gw_samplers.py
```diff
@@ -616,6 +757,52 @@ def _validate_minimum_frequency(
             )
 
 
+def _validate_minimum_frequency_transformer(
```
stephengreen: Should perform certain basic checks, e.g., that `f_min[d] >= domain.f_min`. These are performed in the pre-existing frequency validator.

dingo/gw/inference/gw_samplers.py
```diff
@@ -671,12 +858,151 @@ def check_frequency_updates(
...
+                "Model was not trained with variable frequency ranges "
+                "(no random_strain_cropping, mask_frequency_edges, or mask_random_tokens). "
+                "Cannot update minimum_frequency."
+            )
+        if crop_settings is not None:
+            _validate_minimum_frequency(f_min, detectors, domain, crop_settings)
+        if mask_frequency_edges_settings is not None:
+            _validate_minimum_frequency_transformer(
+                f_min, detectors, domain, mask_frequency_edges_settings
+            )
```
stephengreen, line 796: No validation in case of `"mask_random_tokens"`. Should still perform basic checks.

dingo/gw/inference/gw_samplers.py
stephengreen: More generally, this could perhaps be consolidated with `_validate_minimum_frequency()`.

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1223 @@
...
+        f_min_per_token_single = f_min_per_token[:num_tokens_per_detector]
+        f_max_per_token_single = f_max_per_token[:num_tokens_per_detector]
+
+        mask = np.zeros_like(sample["token_mask"], dtype=bool)
+
+        if self.minimum_frequency is not None:
+            if isinstance(self.minimum_frequency, (float, int)):
+                mask = np.logical_or(
+                    mask,
+                    np.where(f_min_per_token < self.minimum_frequency, True, False),
```
stephengreen: Different behavior from training, see line 644.

dingo/gw/inference/gw_samplers.py
```diff
@@ -309,19 +421,48 @@ def _initialize_transforms(self):
...
+                StrainTokenization(
+                    domain=self.domain,
+                    token_size=tok.get("token_size"),
+                    num_tokens_per_block=tok.get("num_tokens_per_block"),
+                    drop_last_token=tok.get("drop_last_token", False),
+                )
+            )
+            if self.frequency_updates or self.psd_notch_dict:
+                transform_pre.append(
+                    MaskTokensForFrequencyRangeUpdate(
```
stephengreen: Notice that we also apply the zeroing transform `MaskDataForFrequencyRangeUpdate`, and since the token cutoff is off by one with respect to training, it means we see some partially-zeroed tokens.

dingo/gw/inference/gw_samplers.py
```diff
@@ -616,6 +757,52 @@ def _validate_minimum_frequency(
...
+    mask_frequency_edges_settings: dict,
+):
+    """Validate f_min update against the MaskFrequencyEdges training bounds.
+
+    The model was trained to handle f_min values up to f_max_lower, so any
+    requested f_min must not exceed that bound.
+    """
+    if isinstance(f_min, (float, int)):
+        f_min = {d: f_min for d in detectors}
+    cap = mask_frequency_edges_settings.get("f_max_lower", domain.f_max)
```
stephengreen: `f_max_lower` here is the same as `f_min_upper` in the pre-existing validation on `main`, and vice versa. We should really reconcile the meanings.

examples/dingo_t1/train_settings.yaml
```diff
@@ -0,0 +1,135 @@
...
+      p_mask_012_detectors: [0.6, 0.3, 0.1]
+      p_mask_hlv:
+        H1: 0.3
+        L1: 0.3
+        V1: 0.4
+    mask_frequency_edges:
+      p_mask: 0.2
+      f_max_lower: 100.0
+      f_min_upper: 800.0
+      p_lower_upper_both: [0.4, 0.4, 0.2]
```
stephengreen: Can this be moved outside of `tokenization` to match what we already have in `main`?

stephengreen (commented):
Comments on frequency cropping. In general, I recommend aligning more closely with existing code for cropping on `main`.

dingo/pipe/sampling.py
```diff
@@ -65,8 +66,14 @@ def __init__(self, args, unknown_args):
 
         # Frequencies
         # self.sampling_frequency = args.sampling_frequency
-        # self.minimum_frequency = args.minimum_frequency
-        # self.maximum_frequency = args.maximum_frequency
+        self.minimum_frequency = args.minimum_frequency
+        self.maximum_frequency = args.maximum_frequency
```
stephengreen: Why does this get uncommented? Why was it commented before?

dingo/pipe/sampling.py
```diff
@@ -137,6 +144,22 @@ def _load_sampler(self):
 
         self.dingo_sampler.context = self.context
         self.dingo_sampler.event_metadata = self.event_metadata
+        if self.minimum_frequency is not None:
+            self.dingo_sampler.minimum_frequency = self.minimum_frequency_dict
+        if self.maximum_frequency is not None:
+            self.dingo_sampler.maximum_frequency = self.maximum_frequency_dict
```
stephengreen: Not sure why we need this, the frequency bounds should be forwarded via the event metadata.

dingo/pipe/sampling.py
```diff
@@ -137,6 +144,22 @@ def _load_sampler(self):
...
+        # from the stored ASD (covers both the standard and Asimov paths).
+        if self.psd_notch_dict is not None:
+            psd_notch = self.psd_notch_dict
+        else:
+            psd_notch = detect_asd_notches(
+                self.context["asds"], self.dingo_sampler.domain
+            )
+        if psd_notch is not None:
+            logger.info(f"PSD notch detected/configured: {psd_notch}")
+            self.dingo_sampler.psd_notch_dict = psd_notch
```
stephengreen: Should this be in `data_generation.py`, with the information forwarded to the sampler via the event metadata?

dingo/gw/gwutils.py
```diff
@@ -144,3 +144,100 @@ def get_standardization_dict(
...
+
+    Returns
+    -------
+    dict or None
+        ``{det: [[f_lo, f_hi], ...]}`` for each detector that has at least one
+        notch.  Returns ``None`` when no notches are found.
+    """
+    from dingo.gw.noise.asd_dataset import HIGH_ASD_VALUE
+
+    sample_freqs = domain.sample_frequencies  # length max_idx + 1
```
stephengreen, line 215: This assumes the ASD domain and `domain` are the same. However, at the stage this is called, the ASD is always on a uniform grid; but `domain` could be MFD.

dingo/gw/gwutils.py
```diff
@@ -144,3 +144,100 @@ def get_standardization_dict(
...
+        valid_freqs = sample_freqs[min_idx:]
+
+        is_notched = valid_asd >= HIGH_ASD_VALUE * 0.5
+
+        if not np.any(is_notched):
+            continue
+
+        changes = np.diff(is_notched.astype(int))
+        starts = list(np.where(changes == 1)[0] + 1)
+        ends = list(np.where(changes == -1)[0] + 1)
```
stephengreen, line 230: Why are we working with ints instead of bools?

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1223 @@
...
+            # Last token is incomplete: extrapolate f_max
+            if isinstance(domain, MultibandedFrequencyDomain):
+                last_delta_f = domain.delta_f[-1]
+            else:
+                last_delta_f = domain.delta_f
+            f_max_pad = (
+                self.f_max_per_token[-1] + self.num_bins_per_token * last_delta_f
+            )
+            self.f_max_per_token = np.append(self.f_max_per_token, f_max_pad)
+            self.num_padded_f_bins = int((f_max_pad - freqs[-1]) / last_delta_f)
```
stephengreen: Better to use integer arithmetic.

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1223 @@
...
+                mask_detectors = np.logical_or(
+                    input_sample["token_mask"][mask_mod], mask_detectors
+                )
+                # Update mask
+                input_sample["token_mask"][mask_mod] = mask_detectors
+
+        return input_sample
+
+
+class MaskFrequencyEdges(object):
```
stephengreen: The masking logic in these various classes is rather complex, and somewhat repeated. Can it be simplified or unified?

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1223 @@
...
+    * Samples f_upper from [f_lower, f_lower + max_width].
+    * Converts f_lower and f_upper to tokens and creates a token mask removing all tokens in [f_lower, f_upper].
+    """
+
+    def __init__(
+        self,
+        domain: UniformFrequencyDomain | MultibandedFrequencyDomain,
+        p_per_detector: float,
+        f_min: float,
+        f_max: float,
```
stephengreen: Better to default to the whole domain.

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1223 @@
...
+        )
+        f_lower = np.where(mask_interval, f_lower_full, np.inf)
+
+        # f_upper from [f_lower, f_lower + interval_max_width]
+        # Sampling f_upper depends on the f_lower sampled for each (batch, detector).
+        # np.apply_along_axis(np.argwhere, ...) requires that all (batch, detector)
+        # combinations produce the same number of valid upper frequencies.
+        # This is guaranteed: f_values_base_domain is always uniformly spaced (fixed delta_f),
+        # so any window [f_lower, f_lower + max_width] with f_lower on the grid contains
+        # exactly floor(max_width / delta_f) + 1 frequencies. For MFD we use the base domain,
```
stephengreen: Is this true?

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1223 @@
...
+        # f_lower from [interval_f_min, interval_f_max - interval_max_width]
+        mask_f_vals_lower = np.logical_and(
+            self.interval_f_min <= f_values_base_domain,
+            f_values_base_domain <= self.interval_f_max - self.interval_max_width,
+        )
+        possible_f_vals_lower = f_values_base_domain[mask_f_vals_lower]
+        f_lower_full = np.random.choice(
+            possible_f_vals_lower, replace=True, size=batch_block_size
+        )
+        f_lower = np.where(mask_interval, f_lower_full, np.inf)
```
stephengreen: What's happening here with the `np.inf`?

stephengreen (commented):
Further comments

stephengreen:
I've pushed a stack of fix commits on top of this branch, one per topic, addressing the main points of my review:

- **Loads dingo-t1 networks.** Old settings are converted at load time, so the published Dingo-T1 network now runs here.
- **`hidden_features` restored on `DenseResidualNet`.** glasflow scales the spline parameters by it when present; without it the same weights give a different flow, i.e. wrong posteriors. ⚠️ Any network trained on *this* branch with `layer_norm: True` before this fix needs retraining.
- **Detector validation** fixed (the `p_mask_hlv` check was inverted), and `event_metadata` no longer drops `detectors`.
- **Frequency-range validation** unified between the sampler and `dingo_pipe`; partial per-detector dicts now work.
- **Crashes fixed** for non-power-of-two segment durations and for notch detection on multibanded models.
- **Settings renamed** to match `random_strain_cropping`: `mask_frequency_edges` → `mask_frequency_range` (bound names follow cropping's convention), `mask_frequency_interval` → `mask_frequency_notches`. The converter maps the dingo-t1 names; the interim names from this branch aren't supported.
- **PSD notches** are detected at data generation and recorded in the event file, like the frequency bounds.
- Bin-level frequency masking is no longer applied for tokenized models (the token mask already covers it).

Checked end to end against dingo-t1 on GW190701 with the published network: same posteriors and evidence once `hidden_features` is restored.

Still to come: example/docs fixes and some cleanup. I'll raise a few design questions separately (detector indexing beyond HLV, the NaN-PSD handling dingo-t1 had in data generation, average pooling with all tokens masked).

examples/dingo_t1/train_settings.yaml
```diff
@@ -0,0 +1,136 @@
+data:
+  waveform_dataset_path: /01_waveform_dataset/waveform_dataset.hdf5
+  train_fraction: 0.95
+  window:
+    type: tukey
+    f_s: 4096
+    T: 8.0
+    roll_off: 0.4
```
stephengreen: Drop window information from train config.

dingo/gw/transforms/tokenization_transforms.py
```diff
@@ -0,0 +1,1200 @@
...
+        num_blocks: int
+            Number of blocks (= detectors) in GW use case.
+        p_mask_012_detectors: list[float]
+            Specifies the categorical probability distribution for how many detectors to mask, in ascending order.
+            example: [0.1, 0.6, 0.3] = [10% probability to mask 0 detectors (=3 detector setup), 60 % probability for
+            2 detector setup, 30% probability for 1 detector setup]
+        p_mask_hlv: dict
+            Specifies the categorical probability distribution for which specific detectors to mask, order: H1, L1, V1.
+            example: {'H1': 0.1, 'L1': 0.2, 'V1': 0.7} = 10 % probability to mask H1, 20 % probability to mask L1,
+            70% probability to mask V1
```
stephengreen, line 340: We shouldn't hard-code that we have HLV detectors.

examples/dingo_t1/GW150914.ini
```diff
@@ -0,0 +1,88 @@
...
+
+
+################################################################################
+## Data generation arguments
+################################################################################
+
+trigger-time = GW150914
+label = GW150914
+outdir = outdir_GW150914
+channel-dict = {H1:GWOSC, L1:GWOSC, V1:GWOSC}
```
stephengreen: No V1 detector data for this event.

dingo/pipe/data_generation.py
```diff
@@ -410,6 +439,8 @@ def save_hdf5(self):
             "minimum_frequency": self.minimum_frequency_dict,
             "maximum_frequency": self.maximum_frequency_dict,
         }
+        if detected_notches is not None:
+            settings["psd_notch_dict"] = detected_notches
```
stephengreen, line 443: Validate somewhere that the notch dict is consistent with what the network was trained on.

stephengreen: Either raise or warn

dingo/core/nn/transformer.py
```diff
@@ -0,0 +1,320 @@
...
+        num_blocks: int,
+        dropout: float = 0.0,
+        batch_norm: bool = False,
+        layer_norm: bool = False,
+    ):
+        """
+        Parameters
+        ----------
+        input_dims : List[int]
+            [num_tokens, num_features], i.e., the shape of the tokenized waveform,
```
stephengreen: tokenizer acts per-token, why does it need to know the number of tokens?

examples/dingo_t1/train_settings.yaml
```diff
@@ -0,0 +1,129 @@
...
+    transformer_kwargs:
+      d_model: 1024
+      dim_feedforward: 2048
+      nhead: 16
+      dropout: 0.0
+      num_layers: 8
+      norm_first: True
+    pooling: cls
+    final_net_kwargs:
+      activation: elu
```
stephengreen: The final layer shouldn't have activation here. Add a default of no activation, but keep the option. That way the existing demo network still runs.

stephengreen:
Second stack pushed to `transformer` (`bfe1c71b`..`2484a1c7`), on top of the fixes summarized above. All bit-exact against the published Dingo-T1 network (identical transform output and log-probabilities on fixed data), full suite green.

**PSD notching validation** (`bfe1c71b`)
- `psd_notch_dict` is now validated like the frequency bounds: at DAG build on the ini dict and in the sampler setter, via `_validate_psd_notches` / `check_psd_notches`. Configuration errors raise (untrained detector, empty interval, interval touching f_min/f_max). Distribution mismatch warns rather than raises (no `mask_frequency_notches` training, `mask_random_tokens` only, or outside the trained range / `max_width`), since the likelihood stays exact and the network is only a worse proposal; the IS efficiency is the diagnostic. A ResNet model on a notched ASD sees zeros in both whitened strain and inverse ASD, so it gets the same warning.
- `detect_asd_notches` now skips runs touching f_max as well as f_min. bilby fills frequencies beyond a PSD file with `inf`, and the LVK release PSDs (checked GWTC-1, GWTC-4.0, GWTC-5.0) all end one bin short of the round f_max, so every production run produced a spurious one-bin notch there. An edge run wider than one bin warns that the PSD does not cover the model band. Real spectral lines are not an issue: the loudest BayesWave line in those files is ~1e-19 in ASD against the 0.5 threshold.
- `_check_inference_masking_distribution` deleted (its budget had no notch allowance, so only an in-distribution notch could trigger it); docs updated accordingly.

**Cleanup** (`b91cdd7f`, `f48f42b7`, `fe8883d3`, `2484a1c7`)
- The event_metadata setter rebuilt the transform chain once per setting (up to five times per event); it rebuilds once now. Per-batch prints in `MaskTokensForFrequencyRangeUpdate.__call__` removed.
- `MaskFrequencyNotches` applies the per-detector Bernoulli mask directly instead of ±inf/−1 sentinel bounds (seeded comparison bit-identical).
- Tests: a real `_run_sampler` test, chi-square uniformity checks in place of `atol=5, rtol=5`, duplicates removed, `detect_asd_notches` tests moved next to the function.
- `Tokenizer(input_dims=[num_tokens, num_features])` → `Tokenizer(input_dim)`; only the feature count was ever used. Autocompleted from the sample in `build_model`, and `update_model_config` maps the stored `input_dims` of older checkpoints.

**One thing to flag for future trainings:** the one-layer final net (`LinearLayer`) applies its activation *after* the projection, so with `activation: elu` the flow context is bounded below by −1. I've made the activation optional (a bare projection when omitted) and dropped it from the example settings; networks with a stored activation, including the published one, are unaffected. I'd expect neutral-to-slightly-better performance from the bare projection, but haven't measured it.

Full findings/verification record stays in the review notes; the remaining design questions (`DETECTOR_DICT` indexing, augmentation schema, NaN PSDs, position normalization, average-pooling NaN, fastpath toggle, batch_norm on 3-D input, FMPE/GNPE failure modes, `layer_norm` interface with #358) I'll raise separately.

stephengreen:
Third commit pushed to `transformer` (`570dfb32`): detector identity comes from the training detector list rather than a global table. **This one changes the training settings schema**, so please read the first section if you have settings files or networks on this branch.

**Settings change: `tokenization.mask_detectors`**
- `num_blocks` is gone (it is the length of `data.detectors`, which the builder now passes).
- `p_mask_012_detectors` → `p_num_masked`: the categorical over how many detectors to mask, entry k = probability of masking k, length = number of detectors.
- `p_mask_hlv` → `p_detector`: the categorical over which detector to mask, keyed by name; drawn without replacement when more than one is masked. The old name and defaults assumed H1/L1/V1.
- Example yaml and the docs snippet are updated. Old settings inside saved networks are converted automatically at load (both the dingo-t1 `drop_detectors` keys and this branch's `p_mask_*` keys), so nothing needs retraining for the rename alone. Training settings *files* need the new keys.

**What changed underneath**
- The token detector index is now the detector's position in the training `data.detectors` list; `DETECTOR_DICT` is deleted. Any detector set and order trains (LV, HK, ET, …); previously only prefixes of H1, L1, V1 worked, and `[L1, V1]` failed with a `KeyError` in the `MaskDetectors` defaults.
- `StrainTokenization` takes the detector list in block order and, at inference, the training list, and maps event detectors to their training positions. It no longer reads detector order from the ASD dict. That was a latent bug: the dict comes back from the event file in alphabetical order while the strain blocks follow the settings list, so a non-alphabetical `detectors` order mislabeled tokens silently. HLV is alphabetical, which is why it never showed.
- `MaskTokensForFrequencyRangeUpdate` inverts indices with the training list; `build_model`'s one-hot width still comes from the sample and now equals the list length.

**Networks trained on this branch before this commit**
- The published Dingo-T1 network stores `[H1, L1, V1]`, whose list positions coincide with the old table, and converts unchanged: transform output and log-probabilities are bit-identical to before.
- Any old-schema network whose detector list is not a prefix of H1, L1, V1 is now refused at load with a clear error, since list indexing would silently permute its one-hot. I don't believe any exist.

**Verification**
- Bit-exact T1 inference chain; seeded `MaskDetectors` draws identical to the old implementation.
- A real two-epoch `dingo_train` on a tiny transformer with `detectors: [V1, H1]` and all four mask transforms, followed by inference on `[H1]`, `[V1]`, `[H1, V1]` from the saved network.
- New `tests/gw/test_train_builders.py`: builds the full training transform chain from settings on an in-memory dataset, autocompletes and constructs the model, and runs one loss/backward pass, so wiring mistakes in the training path fail in the default suite.

Left as is for now: the repackaging transform still doesn't stamp its detector order onto the sample (the tokenizer is told the same list at construction instead), and the strict `>` upper-edge rule masks a zero-padded last token when the upper bound is the domain edge, which predates this PR's review.

stephengreen:
@annalena-k @nihargupte-ph — a question on NaN handling in `dingo_pipe` data generation, which I'd like to settle before merging rather than fix on my own.

On the `dingo-t1` branch, `data_generation` checked each interferometer's ASD after bilby_pipe's estimate and, if it contained NaNs, either re-estimated the PSD through Dingo's `download_psd` (which rolls the segment back to a NaN-free one) when `--shift-segment-for-psd-generation-if-nan` was set, or raised. This branch is a rewrite off `main` and has neither: an ASD estimated from a segment with NaNs is written to the event file as is, nothing downstream checks it, and the failure surfaces later as NaNs in the whitened data.

Three things I'd like to know:

1. Annalena, how often did this actually bite? Presumably specific O3/O4 events with gaps in the PSD segment. Was the roll-back essential for the paper's events, or a convenience?
2. Is this transformer-specific in any way, or a general `dingo_pipe` issue that should go to `main` as its own small PR? My reading is the latter.
3. Preferred form: at minimum I'd raise immediately on a non-finite ASD with a clear message (cheap, and it turns a silent failure into a loud one). Whether the flag-gated roll-back should come back, and whether it belongs in `dingo_pipe` or in the ini's PSD settings, I'd leave to whoever owns that path.

If we agree on "raise now, roll-back later on `main`", I can add the check here in a few lines.

nihargupte-ph:
> @annalena-k @nihargupte-ph — a question on NaN handling in `dingo_pipe` data generation, which I'd like to settle before merging rather than fix on my own.
> 
> On the `dingo-t1` branch, `data_generation` checked each interferometer's ASD after bilby_pipe's estimate and, if it contained NaNs, either re-estimated the PSD through Dingo's `download_psd` (which rolls the segment back to a NaN-free one) when `--shift-segment-for-psd-generation-if-nan` was set, or raised. This branch is a rewrite off `main` and has neither: an ASD estimated from a segment with NaNs is written to the event file as is, nothing downstream checks it, and the failure surfaces later as NaNs in the whitened data.
> 
> Three things I'd like to know:
> 
> 1. Annalena, how often did this actually bite? Presumably specific O3/O4 events with gaps in the PSD segment. Was the roll-back essential for the paper's events, or a convenience?
> 2. Is this transformer-specific in any way, or a general `dingo_pipe` issue that should go to `main` as its own small PR? My reading is the latter.
> 3. Preferred form: at minimum I'd raise immediately on a non-finite ASD with a clear message (cheap, and it turns a silent failure into a loud one). Whether the flag-gated roll-back should come back, and whether it belongs in `dingo_pipe` or in the ini's PSD settings, I'd leave to whoever owns that path.
> 
> If we agree on "raise now, roll-back later on `main`", I can add the check here in a few lines.

1). I recall out of 50 event analyzed in O3, 5 of them had this issue. But I think in production this will not happen because one points directly to the PSD paths on CIT which will be handled by asimov. 

2). I don't think this is transformer specific. 

3). I think raise now roll-back later on main makes sense. So I guess this means that you would have a parser setting that explicitly tells `dingo_pipe` to roll back the PSD if there is a NaN? EG error out if there is a NaN unless that parser setting is turned on in which case it will roll back the segment and estimate the PSD from that?

stephengreen:
@max-dax — a training-dynamics question on the transformer tokenizer that I'd like your view on before anyone retrains. Token positions (f_min, f_max in Hz, up to 1810 Hz for the MFD) enter the tokenizer's residual blocks only through the GLU gates, together with the detector one-hot. With default initialization the gate pre-activations are O(100) (median |z| ≈ 110 at the example size), so about two thirds of the gate units have exactly zero sigmoid derivative at init and never train; the detector one-hot (±0.45 at most) is invisible to those units. The published Dingo-T1 networks were trained this way and hit the paper's numbers, so it clearly works, but it may cost convergence and positional/detector resolution. This branch now has a `tokenization.normalize_position` option (Hz → [0, 1] over the training band, applied after all mask transforms; default `false` for consistency with the trained network). Do you have intuition or prior experiments on whether normalized positions help here, and would you default it to `true` for new trainings?

(Option added in `9696372c`.)

stephengreen:
Thanks. Agreed on all three. Landed here as "raise now" (`fe81b16f`): data generation raises if an ASD contains NaN within the model band, with a pointer at the PSD settings. Only NaN: inf, which bilby produces beyond a PSD file's range, already whitens to zero and is handled as a notch. Yes on the flag design: on `main`, a parser option would opt into rolling the PSD segment back on NaN strain, and without it the failure stays loud. That follow-up I'd leave to whoever owns the pipe.

max-dax:
> Do you have intuition or prior experiments on whether normalized positions help here, and would you default it to `true` for new trainings?

Yes, I think we should use normalised positions. I would also set this as default in the future. I think we should avoid accumulating a bunch of unnatural default options, just to preserve backward compatibility.

stephengreen:
@max-dax Thanks — making it the default. The loader backfills `false` for networks saved without the key, so Dingo-T1 is unchanged.

To put a number on the saturation, I checked the published network directly. Position enters only through the single GLU gate in the tokenizer's residual block, and the token grid is fixed, so the gate can be evaluated on every context the network ever saw. Holding a token's content fixed and moving its position label by one token changes its embedding by 9–29 % of its norm below ~800 Hz, but by 0.1 % above ~1 kHz, the same as at random initialization. Relabeling a token's detector changes the embedding by 15–23 % below 300 Hz and 2–3 % above 1 kHz. So the network did learn position and detector dependence through the gate over most of the band, but the top of the band stayed where initialization left it. Normalized positions put every gate unit in its responsive regime at every frequency from the start.

stephengreen:
@max-dax @nihargupte-ph — summary of everything on `transformer` since the detector-identity commit (`570dfb32`). One settings change, several fixes, and a large simplification of the tokenization transforms. From the review's side everything is now in place; the remaining step is the merge of the sampler revamp (#389 into `hackathon-1`), after which this branch merges `hackathon-1` and its inference path moves onto the factorized sampler.

**Settings change: `tokenization.normalize_position` now defaults to `true`** (`970375d1`, following Max's answer above). Token positions are rescaled from Hz to [0, 1] before they enter the embedding; in Hz, about two thirds of the GLU gate units sit at zero gradient at initialization. Training writes the resolved value into the saved settings, and the loader backfills `false` for any network saved before the key was recorded, so the published Dingo-T1 network and any interim trainings on this branch are read as trained in Hz and are bit-identical at inference. The example yaml and docs carry the option explicitly.

**Fixes**
- `326ec909` FMPE with the transformer embedding is wired through `create_cf`; GNPE with tokenization raises "not yet supported" in both the training builder and the GNPE sampler rather than failing on the first batch.
- `05ff8f58` The transformer fast-path switch happens once at module import instead of per constructor. The fused kernel crashes on B200; the inference-speed cost is accepted.
- `fe81b16f` `dingo_pipe` data generation raises if an ASD contains NaN within the model band, pointing at the PSD settings (discussed with Nihar above; a roll-back option is a `main`-side follow-up).
- `9696372c` `average` pooling clamps its denominator, so a fully masked sample embeds to a finite value instead of NaN; `batch_norm=True` in the embedding is refused, since `BatchNorm1d` would normalize over the token axis.
- `5930a585` `Tokenizer` is renamed `TokenEmbedding` and takes a generic position layout, `position_continuous_dim` plus `position_category_sizes`, instead of `num_blocks`; leading continuous columns pass through, trailing categorical columns are one-hot. Saved networks convert at load, and the state-dict and metadata keys are unchanged, so Dingo-T1 loads bit-exactly.

**Tokenization transforms simplified** (`1a47893f`, `108ce72e`, `35f41046`; 1185 → 874 lines, same behavior)
- `MaskFrequencyRange` was two copies of one algorithm, for a cut per detector and a cut shared by all detectors. It is now one implementation that draws in the order the two branches drew, so masks and random stream are bit-identical for disjoint cut intervals, verified by a seeded A/B against the previous code on uniform and multibanded domains, batched and unbatched. The zero-information fallback, which keeps only one cut when both together would mask every token, now always applies instead of being gated on interval overlap. For overlapping intervals, as in the published network's settings, the mask distribution is unchanged but a given seed no longer reproduces the old masks, because the old code drew the fallback choice eagerly.
- `MaskTokensForFrequencyRangeUpdate` stores its bounds per detector at construction and its call is one loop; the notch and range masks compute their candidate grids once. Batch handling reads the same way in every transform. The Dingo-T1 inference chain and seeded draws of every mask transform are bit-identical.

**Verification.** Full suite green. Seeded A/Bs for every training-side transform against the code before each change, the stored Dingo-T1 inference chain after each commit. Before this branch is declared final I will run a short seeded training on the example settings against Annalena's last commit (`a5a44172`) and account for any difference.

**Deferred to follow-ups after the merge:** context parameters for the transformer (concatenated after pooling, a small change once inference is on the spine); the legacy `Sampler.log_prob` path on tokenized inputs, which the spine port removes; the `layer_norm` interface shared with #358.

stephengreen:
**hackathon-1 merged into this branch (2026-09-11).** With #389 in, `transformer` now sits on top of hackathon-1 and the PR is mergeable again. Three commits:

- `7e2e4c7d` **Merge hackathon-1 into transformer.** Network side resolved on #358's `norm` selector (`"BatchNorm"` / `"LayerNorm"` / `None`); `TokenEmbedding` takes `norm` too, and the settings converter maps the boolean `batch_norm` / `layer_norm` keys of existing checkpoints (the published Dingo-T1 network loads strictly, 318 LayerNorm modules, 0 BatchNorm). The legacy `Sampler` / `GWSampler` classes are gone; this branch's validators (`check_frequency_updates`, `check_psd_notches`, `check_detector_update`) moved to `dingo/gw/frequency_updates.py`.
- `42b05b63` **Port the transformer inference onto the sampler spine.** `GWSamplerContext.from_model_metadata` builds the tokenized chain (tokens, range and notch masks, position normalization) and hands the network `[waveform, position, token_mask]`; `FlowFactor` accepts a tensor or that list. The per-event settings (`detectors`, `minimum_frequency` / `maximum_frequency`, `psd_notch_dict`) are plain keys of `event_metadata`, as dingo_pipe already writes them, so the setter/property plumbing is gone. Semantics: the event record's detector list governs every per-detector expansion (a float applies to all analyzed detectors, a dict may name any subset of them, a detector outside the list is an error); the same expansion feeds the token mask, the bin mask and the likelihood. GNPE and multi-arm detectors (ET) refuse tokenized networks explicitly. The T1 chain is bit-exact against the pre-merge branch (network inputs and log_prob; H1+L1 subset, per-detector f_min, float f_max, a notch).
- `e4fef72f` docs and the T1 example (`local.num_gpus`).

Full suite: 716 passed. Note the PR description's "Inference-time flexibility" paragraph still describes the old `GWSampler` properties; the current mechanism is documented in `docs/source/inference.md` ("Per-event settings").

@max-dax @nihargupte-ph @annalena-k

stephengreen:
**Follow-up (2026-09-11): review and smoke test of the merged branch; two more commits.**

A fresh full review of `transformer` against `main` (spine, transformer, merge and port) found no high-severity issues; the medium/low ones are fixed in `7b90cb0f`:

- the analyzed detectors are taken in training order whatever order the event record lists them (a ResNet's detector blocks are positional; dingo_pipe already reordered, the Python API did not);
- the likelihood is built on the analyzed detectors only, and `Result.interferometers` agrees with it;
- `Result` reconstruction runs the settings converters, so result files written by the dingo-t1 branch load on this branch (@annalena-k: relevant for your existing runs);
- `psd_notch_dict` intervals are accepted as lists or arrays (#415, the HDF5 round-trip case);
- training refuses `tokenization` together with `random_strain_cropping`, and a tokenized network is refused the cropping license at inference: variable frequency ranges for a transformer come from `mask_frequency_range` (token masks), and cropping-only training would never have produced the masks inference applies;
- the dead single-detector `detectors` fix in `Result.__init__` and the tokenization banners on every `Result` load are removed.

`9429a28e` corrects the T1 example's `num_workers` (the value is the total across GPUs, divided across ranks; the old comment said per GPU).

An end-to-end smoke test on CPU (tiny NPE and transformer networks trained from settings, inference on the spine for both, per-event detector subset / f_min / notch on the transformer, Result save and reload, importance sampling, resume from checkpoint) passed at every stage. Full suite: 723 passed. The T1 chain stays bit-exact against the pre-merge branch.

Filed separately: #416 (`dingo_ls` on legacy HDF5 files, pre-existing on main) and #417 (triangular detectors: frequency bounds keyed `ET` vs arm-keyed data, pre-existing on main). Docs follow-ups noted for a later pass: the T1 example page's `maximum-frequency = {H1: 1024, L1: 512}` is below that network's `f_max_lower: 800`, and two pages state the range licenses more narrowly than the code.

PR #406, opened 2026-09-09
LVK pipe fixes

AlexandreGoettel:
This is WIP because I have some bugs to fix in importance_sampling.py, so that I can get a end-to-end asimov test fully working.

However, I am creating this PR so that you can review the overall idea. This might be very useful because I find it difficult to keep an overview of all the different places in the code the settings go through, so I wouldn't be surprised if I missed something.

I basically tried to keep it simple as I am treating this as a hotfix. Note the the PR is against lvk-stable (which is based off the 0.9.8) release so we can make 0.9.9 without having to review IR1 code.

I modify:
- dingo.ini (to bypass the whitespace issue while keeping asimov conventions)
- pipe/main.py to pass more frequency information to the nodes
- pipe/dag_creator.py to use more frequency information when deciding to make IS data. Basically, afaik, it should only create a new event_data.hdf5 file if it needs to change the domain in a way that can't be handled with masking.
- pipe/data_generation.py to be "allowed" to make data with other frequency ranges.
- gw/result.py so that rebuild_domain can be used beyond the original base_domain

### DIFF
--- a/dingo/asimov/dingo.ini
+++ b/dingo/asimov/dingo.ini
@@ -96,7 +96,15 @@ maximum-frequency={
 {%- endif %}
 tukey-roll-off={{ likelihood['roll off time'] | default: 1.0 }}
 {% if data contains 'importance sampling updates' -%}
-importance-sampling-updates={{ data['importance sampling updates'] }}
+{%- assign p = data['importance sampling updates'] %}
+importance-sampling-updates={
+   {%- if p contains "maximum frequency" -%}
+   maximum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["maximum frequency"][ifo]}},{%- endfor -%}}}
+   {%- endif -%}
+   {%- if p contains "minimum frequency" -%}
+   minimum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["minimum frequency"][ifo]}},{%- endfor -%}}}
+   {%- endif -%}
+}
 {%- endif %}
 
 ################################################################################
--- a/dingo/asimov/dingo.ini
+++ b/dingo/asimov/dingo.ini
@@ -99,10 +99,10 @@ tukey-roll-off={{ likelihood['roll off time'] | default: 1.0 }}
 {%- assign p = data['importance sampling updates'] %}
 importance-sampling-updates={
    {%- if p contains "maximum frequency" -%}
-   maximum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["maximum frequency"][ifo]}},{%- endfor -%}}}
+   maximum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["maximum frequency"][ifo]}},{%- endfor -%}}},
    {%- endif -%}
    {%- if p contains "minimum frequency" -%}
-   minimum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["minimum frequency"][ifo]}},{%- endfor -%}}}
+   minimum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["minimum frequency"][ifo]}},{%- endfor -%}}},
    {%- endif -%}
 }
 {%- endif %}
--- a/dingo/asimov/dingo.ini
+++ b/dingo/asimov/dingo.ini
@@ -98,11 +98,11 @@ tukey-roll-off={{ likelihood['roll off time'] | default: 1.0 }}
 {% if data contains 'importance sampling updates' -%}
 {%- assign p = data['importance sampling updates'] %}
 importance-sampling-updates={
-   {%- if p contains "maximum frequency" -%}
-   maximum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["maximum frequency"][ifo]}},{%- endfor -%}}},
+   {%- if "maximum frequency" in p -%}
+   maximum_frequency: { {%- for ifo in ifos -%}{{ifo}}: {{p["maximum frequency"][ifo]}}{% if not loop.last %},{% endif %}{%- endfor -%} },
    {%- endif -%}
-   {%- if p contains "minimum frequency" -%}
-   minimum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["minimum frequency"][ifo]}},{%- endfor -%}}},
+   {%- if "minimum frequency" in p -%}
+   minimum_frequency: { {%- for ifo in ifos -%}{{ifo}}: {{p["minimum frequency"][ifo]}}{% if not loop.last %},{% endif %}{%- endfor -%} }
    {%- endif -%}
 }
 {%- endif %}
--- a/dingo/gw/result.py
+++ b/dingo/gw/result.py
@@ -180,71 +180,48 @@ class Result(CoreResult):
         which is expected to be populated by reset_event()."""
         updates = self.importance_sampling_metadata["updates"].copy()
 
-        # Assume that updates can contain T, f_s, roll_off, f_min, f_max, but no other
-        # quantities that define a new domain (e.g., delta_f). Typical event metadata
-        # will be constructed in this way.
-
         domain_keys = ["minimum_frequency", "maximum_frequency", "T"]
         if any(k in updates for k in domain_keys):
-            # Handle dicts and floats
-            f_min = updates.get("minimum_frequency", self.domain.f_min)
-            if isinstance(f_min, dict):
-                f_min = min(f_min.values())
-            f_max = updates.get("maximum_frequency", self.domain.f_max)
-            if isinstance(f_max, dict):
-                f_max = max(f_max.values())
-            if "T" in updates:
-                updates["delta_f"] = 1.0 / updates["T"]
+            self.importance_sampling_metadata["use_base_domain"] = True
+            print("Domain will be rebuilt, forcing use_base_domain=True")
+
+        f_min = updates.get("minimum_frequency", self.domain.f_min)
+        if isinstance(f_min, dict):
+            f_min = min(f_min.values())
+        f_max = updates.get("maximum_frequency", self.domain.f_max)
+        if isinstance(f_max, dict):
+            f_max = min(f_max.values())
+        if "T" in updates:
+            updates["delta_f"] = 1.0 / updates["T"]
+        if isinstance(self.domain, MultibandedFrequencyDomain):
+            delta_f = updates.get("delta_f", self.domain.base_domain.delta_f)
+        else:
+            delta_f = updates.get("delta_f", self.domain.delta_f)
+
+        domain_dict_update = {"f_min": f_min, "f_max": f_max, "delta_f": delta_f}
+        domain_dict = self.domain.domain_dict
+        try:
+            self.domain.update(domain_dict_update)
+        except ValueError:
             if isinstance(self.domain, MultibandedFrequencyDomain):
-                delta_f = updates.get("delta_f", self.domain.base_domain.delta_f)
+                base_domain_dict = domain_dict["base_domain"].copy()
+                for k, v in domain_dict_update.items():
+                    base_domain_dict[k] = v
+                domain_dict["base_domain"] = base_domain_dict
+                domain_dict["delta_f_initial"] = delta_f
             else:
-                delta_f = updates.get("delta_f", self.domain.delta_f)
+                domain_dict.update(domain_dict_update)
 
-            domain_update_dict = {}
-            if f_min != self.domain.f_min:
-                domain_update_dict["f_min"] = f_min
-            if f_max != self.domain.f_max:
-                domain_update_dict["f_max"] = f_max
-            if "delta_f" in updates and updates["delta_f"] != delta_f:
-                domain_update_dict["delta_f"] = delta_f
-
-            # Try to update existing domain first
-            try:
-                if domain_update_dict:
-                    self.domain.update(domain_update_dict)
-            except ValueError:
-                domain_dict = self.domain.domain_dict  # Existing settings
-                if isinstance(self.domain, MultibandedFrequencyDomain):
-                    base_domain_dict = domain_dict["base_domain"].copy()
-                    if (
-                        "minimum_frequency" in updates
-                        or "maximum_frequency" in updates
-                        or "T" in updates
-                    ):
-                        base_domain_dict["f_min"] = f_min
-                        base_domain_dict["f_max"] = f_max
-                        if "T" in updates:
-                            base_domain_dict["delta_f"] = domain_dict[
-                                "delta_f_initial"
-                            ] = (1.0 / updates["T"])
-                        domain_dict["base_domain"] = base_domain_dict
-                else:
-                    domain_dict.update(
-                        (k, updates[k]) for k in set(domain_dict).intersection(updates)
+            if verbose:
+                print("Rebuilding domain as follows:")
+                print(
+                    yaml.dump(
+                        domain_dict,
+                        default_flow_style=False,
+                        sort_keys=False,
                     )
-
-                if verbose:
-                    print("Rebuilding domain as follows:")
-                    print(
-                        yaml.dump(
-                            domain_dict,
-                            default_flow_style=False,
-                            sort_keys=False,
-                        )
-                    )
-                self.domain = build_domain(domain_dict)
-        elif verbose:
-            print("No domain updates found; domain not rebuilt.")
+                )
+            self.domain = build_domain(domain_dict)
 
     def _build_prior(self):
         """Build the prior based on model metadata. Called by __init__()."""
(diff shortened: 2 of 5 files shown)


dingo/pipe/main.py
```diff
@@ -280,6 +281,7 @@ def __init__(
         self.prior_dict = args.prior_dict
         self.default_prior = "BBHPriorDict"
         self.minimum_frequency = args.minimum_frequency
+        self.sampling_frequency = args.sampling_frequency
```
nihargupte-ph, line 284: hm what is the need for sampling frequency as a new argument? Shouldn't the importance sampling fmax be determined by what's in the importance sampling updates?

nihargupte-ph:
Note that  IS event data is still truncated to 896 Hz. So save_hdf5 still calls build_domain_from_model_metadata(model.metadata, base=True) which applies domain_update before taking the base domain, and the multibanded _set_new_range also narrows base_domain to 896. The strain is then cut to that domain. I think this needs to be fixed.

nihargupte-ph:
I think that `_rebuild_domain` does not widen the domain because the domain has already been updated once.

dingo/asimov/dingo.ini
```diff
@@ -96,7 +96,15 @@ maximum-frequency={
...
-importance-sampling-updates={{ data['importance sampling updates'] }}
+{%- assign p = data['importance sampling updates'] %}
+importance-sampling-updates={
+   {%- if p contains "maximum frequency" -%}
+   maximum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["maximum frequency"][ifo]}},{%- endfor -%}}}
+   {%- endif -%}
+   {%- if p contains "minimum frequency" -%}
+   minimum_frequency: {{{%- for ifo in ifos -%}{{ifo}}:{{p["minimum frequency"][ifo]}},{%- endfor -%}}}
+   {%- endif -%}
+}
```
nihargupte-ph, line 107: I think this would fail if there are multiple updates because there needs to be a comma in between

dingo/gw/result.py
```diff
@@ -180,23 +180,37 @@ def _rebuild_domain(self, verbose=False):
...
+        else:
+            delta_f = updates.get("delta_f", self.domain.delta_f)
 
-            domain_dict = self.domain.domain_dict  # Existing settings
-            domain_dict.update(
-                (k, updates[k]) for k in set(domain_dict).intersection(updates)
-            )
+        domain_dict_update = {"f_min": f_min, "f_max": f_max, "delta_f": delta_f}
+        domain_dict = self.domain.domain_dict
+        try:
```
nihargupte-ph: Could one instead just define the dict in one and then run .update rather than checking for equality for each entry?

nihargupte-ph:
I think more generally, the likelihood should be derived from the EventDataset rather than the network.

nihargupte-ph:
I think a hotfix (this is definately not a clean permanent solution) could be something like: 

  Edit 1: dingo/pipe/data_generation.py, in save_hdf5

  Replace the single `build_domain_from_model_metadata(..., base=True)` line with:

```
  domain = build_domain_from_model_metadata(model.metadata, base=True)
  if self.importance_sampling:
      # Start from the training domain (before domain_update) so IS data can
      # extend beyond the network's cropped range, then narrow to the request.
      domain = build_domain(model.metadata["dataset_settings"]["domain"])
      domain = getattr(domain, "base_domain", domain)
      domain.update({"f_max": max(self.maximum_frequency_dict.values())})
```

The sampling-time file is untouched. Only f_max matters here, since the array always starts at 0 Hz and per-detector minimum frequencies are applied as ASD masks in the likelihood.

Edit 2: dingo/gw/result.py, replace the body of `_rebuild_domain`

```
  def _rebuild_domain(self, verbose=False):
      updates = self.importance_sampling_metadata["updates"]
      if not any(k in updates for k in ("minimum_frequency", "maximum_frequency")):
          return
      f_min = updates.get("minimum_frequency", self.event_metadata["minimum_frequency"])
      f_max = updates.get("maximum_frequency", self.event_metadata["maximum_frequency"])
      if isinstance(f_min, dict):
          f_min = min(f_min.values())
      if isinstance(f_max, dict):
          f_max = max(f_max.values())
      # Rebuild from the training domain (before domain_update), narrowed to the request.
      self.domain = build_domain(self.base_metadata["dataset_settings"]["domain"])
      self.domain.update({"f_min": f_min, "f_max": f_max})
      if verbose:
          print(f"Rebuilt domain: f_min={self.domain.f_min}, f_max={self.domain.f_max}")
```

that way the domain is only updated once

Also, the multibanded f_max lands one base bin low (1791.875 vs 1792) from the band edge. So we should set `getattr(self.domain, "base_domain", self.domain).f_max` in sample_calibration_parameters.

nihargupte-ph (changes requested):
Hi thanks! I think it's still a draft so I imagine a few things to be changed still but I tried to leave comments as early as I could

stephengreen:
I looked at this alongside Nihar's review of #389, where the same problem came up, and implemented a resolution on `sampler-revamp`: dingo-gw/dingo@06a9d746. Sharing it here because the pipe half applies directly to `lvk-stable`, and the rest maps onto code this branch still has.

**What was wrong.** Two layers. (1) A frequency range under `importance-sampling-updates` never reached the likelihood: data generation popped the keys before generating the IS data (since #287), so the IS event settings carried the network's range, `reset_event` saw no change, and the likelihood masked at the network's band. This has been the case on `main` regardless of the domain rebuild, which for a uniform domain only ever changed `delta_f`. (2) The calibration spline nodes were placed over the domain bounds, in both the marginalization transform and `sample_calibration_parameters`, whereas bilby_pipe places them over each interferometer's analysis range, so they could not match Bilby's whenever the two ranges differed.

**The resolution on `sampler-revamp`** follows one rule: the likelihood is independent of the network and works on the grid of the event data it is given.
- Data generation no longer pops the keys. It generates the IS data on the network's base band extended to the requested range (so widening beyond a `domain_update` band works), at the requested duration, and records that grid in the event settings as `domain`.
- The likelihood builds its grid from that record (files without it are on the network's grid, as before) and extends the waveform-generator domain to contain it. Each detector's range is still applied by ASD masking, and the calibration nodes now span the same per-detector range, as in Bilby.
- A parse-time check requires an IS-time range to be positive, non-empty and at most Nyquist; no strain-cropping license applies, since the network never sees the range. An injection combined with a widened band or a changed duration is refused.
- For a multibanded model the base-domain choice is an argument of the likelihood; the decimated likelihood needs the data on the network's grid.

**Mapping to `lvk-stable`.** The `data_generation.py` and `main.py` changes apply as they are: no pop, grid = network band ∪ requested range at `1/T`, `settings["domain"]`, the parse-time check, and the string form of a per-detector dict for the bilby_pipe setters (which this PR already has). What `sampler-revamp` does in `GWSamplerContext.likelihood()` corresponds to `Result._rebuild_domain` / `_build_likelihood` here: build the likelihood's data domain from `event_metadata["domain"]` (falling back to the network domain), extend the wfg domain to contain it, and leave the masking as it is. That replaces the rebuild, so there is no need to widen an already-updated domain (Nihar's second point) or to force `use_base_domain` from inside it. The calibration-node change applies as it is to `sample_calibration_parameters` and the two transforms.

Two things from this PR I would not carry over: the `min` over per-detector maxima in the rebuild (the grid needs the `max`, with the other detectors masked), and the DAG shortcut that skips IS data generation when masking would do, because the IS stage then receives the sampling-stage event file, `reset_event` records no change, and the range is dropped again.

The branches differ in that `sampler-revamp` no longer has `_rebuild_domain` (the legacy samplers are gone), so the diff will not apply verbatim; the pipe part will.

dingo/asimov/dingo.ini
AlexandreGoettel: Good catch!

dingo/pipe/main.py
AlexandreGoettel: I passed it so I could later test if the new f_max is in Nyquist range, and later decided against it checking against Nyquist (let the user handle it)..
Given other feedback, I re-added a Nyquist check :)

AlexandreGoettel:
Thanks guys!

> I think a hotfix (this is definately not a clean permanent solution) could be something like:
> 
> Edit 1: dingo/pipe/data_generation.py, in save_hdf5
> 
> Replace the single `build_domain_from_model_metadata(..., base=True)` line with:
> 
> ```
>   domain = build_domain_from_model_metadata(model.metadata, base=True)
>   if self.importance_sampling:
>       # Start from the training domain (before domain_update) so IS data can
>       # extend beyond the network's cropped range, then narrow to the request.
>       domain = build_domain(model.metadata["dataset_settings"]["domain"])
>       domain = getattr(domain, "base_domain", domain)
>       domain.update({"f_max": max(self.maximum_frequency_dict.values())})
> ```
> 
> The sampling-time file is untouched. Only f_max matters here, since the array always starts at 0 Hz and per-detector minimum frequencies are applied as ASD masks in the likelihood.
> 
> Edit 2: dingo/gw/result.py, replace the body of `_rebuild_domain`
> 
> ```
>   def _rebuild_domain(self, verbose=False):
>       updates = self.importance_sampling_metadata["updates"]
>       if not any(k in updates for k in ("minimum_frequency", "maximum_frequency")):
>           return
>       f_min = updates.get("minimum_frequency", self.event_metadata["minimum_frequency"])
>       f_max = updates.get("maximum_frequency", self.event_metadata["maximum_frequency"])
>       if isinstance(f_min, dict):
>           f_min = min(f_min.values())
>       if isinstance(f_max, dict):
>           f_max = max(f_max.values())
>       # Rebuild from the training domain (before domain_update), narrowed to the request.
>       self.domain = build_domain(self.base_metadata["dataset_settings"]["domain"])
>       self.domain.update({"f_min": f_min, "f_max": f_max})
>       if verbose:
>           print(f"Rebuilt domain: f_min={self.domain.f_min}, f_max={self.domain.f_max}")
> ```
> 
> that way the domain is only updated once
> 
> Also, the multibanded f_max lands one base bin low (1791.875 vs 1792) from the band edge. So we should set `getattr(self.domain, "base_domain", self.domain).f_max` in sample_calibration_parameters.

I missed the domain_update coming through save_hdf5, however I don't think the above will work because domain.update fails if the new f_max is outside of the original range. Instead, I now make it generate a new domain based off the updated frequency settings.
In build_likelihood, I then simply make it use the information in event_dataset if available, so there is no need for me to change _rebuild_domain anymore.

The good news is I'm ready to remove this from draft stage as I understand the code better, and I have a new end-to-end run going that should hopefully work (takes a while with building a container etc..).

My mindset with this PR is to really only apply a hotfix to urgently get production runs going, as the reviewers are now failing a bunch of DINGO runs, not for this to be used in the long term. With this lvk-stable release, development can still go forward on main etc.

AlexandreGoettel:
I'll also note that Claude's `frequency_updates.py` logic is already more-or-less in my dag_creator.py, as that made more sense to me, but I'm open to any check there.
Also, there are seemingly random updated lines that were forced on me by black :sweat_smile:

stephengreen:
@AlexandreGoettel Thanks. I went through the new commits (head 5bcf0e5b) and I'm afraid they don't get there yet. In order of severity:

1. **Multibanded networks can no longer be importance sampled at all.** `_build_likelihood` now builds a uniform data domain from `event_metadata` whenever it carries `minimum_frequency`, `maximum_frequency` and `T`, which every pipe-generated event file does. For a multibanded network the recorded bound is the base domain's (e.g. 1099.0 Hz) while the waveform-generator domain ends at the last band edge (1098.875 Hz), so `GWSignal._check_domains` raises "Output domain is not contained within WaveformGenerator domain". I checked this by loading a sampling result from a multibanded network with no updates at all: it fails for both `use_base_domain` values. A uniform data domain against a multibanded generator would not work past that check either.
2. **The DAG shortcut drops every update that is not a frequency.** `_importance_sampling_updates_require_new_domain` only looks at `sampling_frequency` and `maximum_frequency`, so `importance-sampling-updates = {'duration': 4.0}` (the example in dingo_pipe.md) or a PSD change no longer regenerates data and is silently ignored. Previously any non-empty update regenerated.
3. **A narrowing request is still ignored.** With `maximum_frequency` at or below the network's, no data are generated, the IS node gets the sampling-stage event file, `reset_event` sees no change, and the likelihood masks at the network's band. That is the original ProdPE bug for any Bilby `f_max` not above the network's.
4. **The main use case crashes at DAG creation.** `assert f_max <= f_s / 2.0` uses the updated sampling frequency, which is `None` when only `maximum_frequency` is given (all the asimov template emits), so `dingo_pipe` raises a `TypeError`.
5. **`min` over per-detector maxima.** The data domain is cut at the smallest detector maximum while the IS data are generated up to the largest, so a per-detector request fails with "Strain data does not match domain". It needs `max`, with the other detectors masked.
6. Smaller: `data_domain = build_domain(...)` sits inside `if isinstance(f_max, dict)`, so a scalar record gives an `UnboundLocalError`; and the calibration nodes still span the domain bounds rather than each detector's range.

For a hotfix on `lvk-stable` I'd take the pipe half of dingo-gw/dingo@06a9d746 as it is: no popping of the keys, no shortcut, IS data generated on the network band extended to the request at the requested duration, and that grid recorded in the event settings as `domain`. In `_build_likelihood`, build the data domain from that record when present (and otherwise keep the current behaviour, in particular for multibanded networks), use `max` over detectors, fall back to the original sampling frequency in the Nyquist check, and dedent the domain construction.

AlexandreGoettel:
> In order of severity:
> 
>     1. **Multibanded networks can no longer be importance sampled at all.** `_build_likelihood` now builds a uniform data domain from `event_metadata` 

When looking at the code in `main`, since domain_update raises a NotImplementedError at multibanded domains, I assumed that the domain at the build_likelihood stage already has to be uniform?

2. & 3. I can fix those for this PR, I only focused on getting the currently urgent GWTC-6 re-runs going for now, which don't need that.

4. & 5. & 6. Goodness, typos all around! Good catch, just fixed that.

AlexandreGoettel:
@nihargupte-ph I just updated the PR with a much simpler version that I tested on a real event and I know works.

VERY MUCH a temporary hotfix! But it will work for the re-runs today.

The logic is as follows:
 - data_generation.py creates two event_data files with different domain information
 - importance_sampling.py manually updates the result objects context and event_metadata to avoid calling the faulty rebuild_domain().
 - result.py: instead, add build_data_domain() function that builds the updated domain. This is then used in _build_likelihood and _sampling_calibration_parameters (the latter was missing from before)

dingo/gw/result.py
```diff
@@ -212,6 +212,35 @@ def _rebuild_domain(self, verbose=False):
...
+                if delta_f != getattr(self.domain,
+                                      "base_domain",
+                                      self.domain).domain_dict["delta_f"]:
+                    raise NotImplementedError("Can't update delta_f")
+
+            data_domain_dict = self.base_metadata["dataset_settings"]["domain"].copy()
+            data_domain = build_domain(data_domain_dict)
+            data_domain.update(domain_dict)
+
+        return data_domain
```
nihargupte-ph, line 240: I think `_get_data_domain` should size the domain from the strain length rather than from `event_metadata["maximum_frequency"]`.

For a strain-cropped run (e.g. `quality: maximum frequency` below the network's) with no `importance-sampling-updates`, no IS event file is generated, so IS reads the sampling-stage file. That file's strain is on the network grid (up to the network f_max, since the network needs that input length), but its settings record the cropped f_max. Building the domain from the recorded value gives a shorter domain than the strain, and the likelihood fails with `operands could not be broadcast together`. This worked on `lvk-stable` via ASD masking.

Taking f_max from the array length covers both kinds of file: IS files are written on the requested range, sampling files on the network grid, and per-detector ranges are still applied by masking. It also handles scalar `minimum_frequency` (as written by `GWSampler` outside dingo_pipe), which currently fails on `.values()`.

I checked this on a multibanded network (base 20-2048 Hz, cropped to 896 Hz) for: IS update to 1792, no updates, cropped f_max 512 without updates, cropped f_min, per-detector f_min, PSD-only update with cropping, and a uniform-domain network, with both `use_base_domain` values. All build the likelihood; without this change the cropped-f_max cases and the scalar case fail.

```suggestion
    def _get_data_domain(self):
        """
        Get the data domain on the grid of the event data in self.context.

        The event file is on the network grid (sampling stage) or on the network band
        narrowed/extended to the requested range (IS stage), so f_max is taken from the
        strain length. Per-detector ranges are still applied by ASD masking.

        Returns a Domain object.
        """
        if self.context is None or self.event_metadata is None:
            return self.domain
        base_domain = getattr(self.domain, "base_domain", self.domain)
        T = self.event_metadata.get("T")
        if T is not None and 1.0 / T != base_domain.delta_f:
            raise NotImplementedError("Can't update delta_f")

        num_bins = len(next(iter(self.context["waveform"].values())))
        f_min = self.minimum_frequency
        if isinstance(f_min, dict):
            f_min = min(f_min.values())

        data_domain = build_domain(self.base_metadata["dataset_settings"]["domain"])
        f_max = (num_bins - 1) * base_domain.delta_f
        data_domain.update({"f_min": f_min, "f_max": f_max})
        return data_domain
```

AlexandreGoettel: Overall that makes sense to me and it works on my test as well :)
It feels less readable to use the self.minimum_frequency property which uses the event_metadata and then infers f_max, instead of just getting both from event_metadata, but if it works for the reruns that's all I want :sweat_smile:

PR #414, opened 2026-09-11
Fix MFD threshold search crash when band count changes

nihargupte-ph:
## Problem

`dingo_generate_multibanded_domain` crashes during the threshold search with

```
File ".../generate_multibanded_domain.py", line 825, in generate_multibanded_domain_settings
    if np.allclose(lo_nodes, hi_nodes):
ValueError: operands could not be broadcast together with shapes (6,) (5,)
```

The search first brackets the target median mismatch between two thresholds, then bisects. Both the bracket check (`np.allclose(lo_nodes, hi_nodes)`) and the bisection convergence check (`np.allclose(mfd.nodes, previous_nodes)`) assume the compared MFDs have the same number of bands. A larger threshold can drop a band, so the node arrays differ in length and `np.allclose` raises instead of returning `False`.

This is easy to hit with token-aligned banding (`--token_size 16`), because bands snap to whole tokens. Reproduced on an 8 s SEOBNRv5PHM dataset (chirp mass 12–150, f_start 9 Hz, 20–2048 Hz, target median mismatch 1e-3). The bracket is [0.32, 5.12], giving 5 vs 4 bands. It crashes with both 1000 and 5000 waveforms.

## Fix

- Add `_same_nodes(nodes_a, nodes_b)`, which returns `False` when the shapes differ and otherwise uses `np.allclose`.
- Use it at both comparison sites.

MFDs with different band counts are by definition different bandings, so the search simply continues bisecting.

## Tests

- **Regression test:** mocks waveform generation so the bracketing walk ends at thresholds giving 3 vs 2 bands. It fails on `hackathon-1` with the broadcast error and passes with the fix.
- **Unit tests for `_same_nodes`:** identical nodes, same length but different nodes, and different band counts.
- `tests/gw/test_multibanded_domain_generation.py`: 41 passed.
- **End to end:** on the real case above with 5000 waveforms, the fixed code converges to nodes `[20, 40, 52, 68, 84, 2036]` (median mismatch 9.0e-4).

https://claude.ai/code/session_012UCuZKieZZ8h8XkutPLLfF

### DIFF
--- a/dingo/gw/dataset/generate_multibanded_domain.py
+++ b/dingo/gw/dataset/generate_multibanded_domain.py
@@ -595,6 +595,25 @@ def _output_settings_path(settings_file: str) -> str:
     return os.path.join(directory, out_name)
 
 
+def _same_nodes(nodes_a: np.ndarray, nodes_b: np.ndarray) -> bool:
+    """Return whether two MFD node arrays describe the same banding.
+
+    Node arrays of MFDs with a different number of bands have different lengths, which
+    ``np.allclose`` cannot broadcast; such MFDs are simply not the same.
+
+    Parameters
+    ----------
+    nodes_a, nodes_b : np.ndarray
+        MFD node arrays.
+
+    Returns
+    -------
+    bool
+        True if both arrays have the same shape and all nodes agree.
+    """
+    return np.shape(nodes_a) == np.shape(nodes_b) and np.allclose(nodes_a, nodes_b)
+
+
 def generate_multibanded_domain_settings(
     settings_file: str,
     num_samples: int,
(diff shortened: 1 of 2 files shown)


stephengreen (approved):
Looks good. The shape check before `np.allclose` is the right fix at both sites, and the search stays bounded by the iteration cap and the bracket-width test. Ran `tests/gw/test_multibanded_domain_generation.py` locally on the PR head: 41 passed.

Two notes, neither blocking: `main` carries the identical file, so the same crash exists there until the next hackathon-1 → main sync. And the pytest workflow only triggers for PRs into `main`, which is why this PR shows only the docs build; adding `hackathon-1` to the workflow's branch list would give these PRs a CI run.

PR #419, opened 2026-09-14
Add tidal waveform support and exact co-rotating synthetic phase

lorenzopompili00:
This PR makes the (2, 2)-approximation synthetic phase **exact** for all approximants with only a co-precessing $(2, \pm 2)$ pair — IMRPhenomPv2, IMRPhenomXP, the NRTidal family — via a new `co_rotate_spins` option, at the same single waveform evaluation per sample. It also adds support for tidal approximants called through LALSimulation (tested with IMRPhenomXP_NRTidalv3).

## Motivation

DINGO-BNS analyses have so far reconstructed the phase with `approximation_22_mode: true` ([e.g. the GW170817 demo config](https://github.com/dingo-gw/binary-neutron-star-demo/blob/main/GW170817/inference-dingo-pipe/GW170817.ini#L18)): one waveform evaluation per sample, assuming the waveform transforms as a global $e^{2i\phi}$. In Dingo's convention, however, phase shifts leave the Cartesian spins fixed (`spin_conversion_phase = 0`), and under that convention the assumption is **not exact for precessing models**: precession spreads the co-precessing $(2, \pm 2)$ pair over all inertial-frame $m \in [-2, 2]$, which breaks the $\phi \to \phi + \pi$ degeneracy the approximation relies on. For a mildly precessing BNS (spins $\sim + 0.05$, IMRPhenomXP_NRTidalv3) the induced waveform mismatch ranges from $\sim 10^{-5}$ near face-on to a few $\times 10^{-2}$ near edge, and grows to order-unity mismatches for larger in-plane spins.                                         

## `co_rotate_spins`: the (2, 2) reconstruction made exact

The fix follows from a convention observation. In bilby's parameterization the `phase` parameter enters the spin conversion (`bilby_to_lalsimulation_spins`), so a phase shift co-rotates the in-plane spins with the orbital phase — a rigid rotation of the source about $\hat{L}$. For models with a single co-precessing $(2, \pm 2)$ pair, that rotation turns the whole waveform into an exact global $e^{2i\phi}$ factor (verified to double precision). This is precisely why bilby's analytic phase marginalization — $\ln I_0(|\langle d|h\rangle|)$ in [`bilby.gw.likelihood`](https://github.com/bilby-dev/bilby/blob/main/bilby/gw/likelihood/base.py), derived for (2, 2)-only signals — is exact for precessing models like IMRPhenomPv2 in bilby: its phase parameter *is* the co-rotating one. Dingo's fixed-spin phase does not have this property, but the two conventions coincide at `phase = 0` and differ only by a rotation of the spin angles, which can be applied after the draw.

Setting `co_rotate_spins: true` in the synthetic-phase settings therefore:

1. draws the phase from the (2, 2) profile, computed from one waveform call at `phase = 0` (unchanged code path) — the exact phase conditional in the co-rotating convention,
2. relabels the sample back into the model convention — `theta_jn` / `phi_jl` rotate; tilts, `phi_12`, and magnitudes are invariant — via the existing `SpinConventionReparam` chain step run in a new `direction="to_network"`,
3. adds the rotation Jacobian $\sin\theta_{JN} / \sin\theta_{JN}'$ to the proposal density (the map preserves the spherical measure).

Eligibility is verified at runtime by a two-waveform probe (no approximant-name matching); on failure — e.g. genuine higher modes — the run warns and falls back to the exact mode sum. For existing BNS workflows this is a drop-in upgrade of `approximation_22_mode: true` at the same cost; compared to the exact mode sum it is equally exact at $\sim 6\times$ less cost (0.23 vs 1.5 s per sample on a 256 s BNS domain). Note the output samples' `theta_jn` / `phi_jl` are relabeled per sample along with the added `phase`.

## Exact mode sum for tidal models (opt-in)

LALSimulation implements no `SimInspiralChooseFDModes` for NRTidal approximants, so the individual-mode path can never serve them; the DFT phase decomposition can, from polarization evaluations. With `mode_list=[[2, 2], [2, -2]]` the reconstruction matches direct regeneration to round-off. There is deliberately no `DEFAULT_ELL_MAX` entry: with `co_rotate_spins` exact at a single evaluation, the 5-evaluation mode sum is opt-in for this model. Related hardening:

- The `generate_FD_modes_LO` error now explains the no-FD-modes case and points at the DFT path and `co_rotate_spins`.
- The DFT inverts waveforms that differ only by a phase shift, so it requires a fixed `spin_conversion_phase`; with `None` a phase shift also rotates the in-plane spins (measured: the two m-decomposition paths then disagree at mismatch $\sim 0.8$ on subdominant components). `generate_hplus_hcross_m` now warns and falls back to the individual-mode path in that case.

## Tidal waveform support

The LAL path already inserts `lambda_1` / `lambda_2`, and everything downstream is dictionary-driven; verified for IMRPhenomXP_NRTidalv3 on uniform and multibanded domains. Added:

- `lambda_1: default` / `lambda_2: default` priors (`Uniform(0, 5000)`).
- LALDict aliasing fix: tidal insertion no longer mutates the shared mode-array dict (stale lambdas could leak into later lambda-less calls).
- bilby-side injections with tidal parameters route to `lal_binary_neutron_star` (parameter-driven; injections via `injection-file` are not inspected).
- Docs: a "Tidal parameters" section in `bns.md` with the required BNS settings, and the `co_rotate_spins` kwarg in `result.md`.

## Validation

New deterministic tests: `tests/gw/waveform_generator/test_wfg_tidal.py` (tidal generation on UFD/MFD with a silent-drop regression guard, DFT reconstruction and aliasing, documented 22-approximation mismatch, LALDict non-mutation, prior defaults, co-rotation probe on an eligible and a higher-mode model), chain-level tests of the co-rotation proposal density and reverse fold (`test_synthetic_phase.py`), the reversed reparametrization (`test_spin_convention_reparam.py`), and the DFT spin-convention fallback (`test_wfg_m.py`). End-to-end, the exact likelihood at co-rotated chain samples matches the drawn profile to the chain's float32 floor.

## Usage

```ini
importance-sampling-settings = {synthetic_phase: {co_rotate_spins: true, n_grid: 5001, uniform_weight: 0.01}}
```


stephengreen:
A note on IMRPhenomXP_NRTidalv3 evaluated on a frequency sequence, found while wiring up multibanded BNS training on top of this PR (#428–#432).

**Observation.** LALSimulation's NRTidalv3 PhenomX wrappers place the waveform's time origin at `min(f_merger, last requested frequency)`. In `LALSimIMRPhenomX.c` (tidal branch of `IMRPhenomXASGenerateFD`): `f_final = freqs->data[freqs->length-1]`, lowered to `f_merger` if that is smaller, then `tshift = -dphi_fmerger` is added to `linb`. So a `SimInspiralChooseFDWaveformSequence` call whose frequencies end below the merger returns a waveform aligned at its last frequency, not at the merger. Measured with dingo's parameter conversion on a 20–256 Hz domain: a 1 Hz grid ending at 255.3 Hz vs one ending at 256.0 Hz differ by a 1.156 ms time shift (Newtonian t(255.3) − t(256) = 1.159 ms); a grid ending at 300 Hz is shifted by −54.1 ms (Newtonian 54.5 ms). `IMRPhenomXAS_NRTidalv3` behaves the same; `IMRPhenomPv2_NRTidal`, `IMRPhenomXP`, `IMRPhenomXPHM`, and `IMRPhenomD` are independent of the requested grid.

**Where it matters in dingo.** The multibanded route evaluates the waveform at the band centers, so the last requested frequency is the last bin center rather than `f_max`. If the dataset domain ends below the merger, the training waveforms are shifted in time relative to base-domain data (injections, real data, likelihood templates), which are generated on the uniform dataset domain by `SimInspiralFD`. On a toy 20–256 Hz domain this gave a 5 % mismatch between the training input and the sampler context's data preparation. With the BNS dataset domain to 2048 Hz (past the merger for the whole prior) the two agree to mismatch < 1e-6, including the BBH-like corner of the prior; `domain_update` and frequency updates at inference do not change the domain templates are generated on, so they are unaffected. The same caveat applies to any other frequency-sequence evaluation ending below the merger with this model.

Suggest mentioning this in the PR (the "verified on multibanded domains" claim holds only for domains extending past the merger) and possibly upstream, since it is a property of the LAL model rather than of dingo. The BNS docs page in #431 carries a warning to the same effect.

lorenzopompili00:
A reminder for myself: the `co_rotate_spins` option would allow to cache the synthetic-phase log likelihood for importance sampling as in https://github.com/dingo-gw/dingo/pull/422, this option should be added if this PR merged after https://github.com/dingo-gw/dingo/pull/422

lorenzopompili00:
@stephengreen Thanks, nice catch! I can reproduce it on this branch. Rather than only documenting this, the PR now works around it in the generator: every `SimInspiralChooseFDWaveformSequence` call asks for one extra frequency at the base domain's `f_max`, and the extra sample is dropped afterwards. The model then places the waveform in time as if it had been generated with `SimInspiralFD`, no matter where the band structure ends. For all other models the results are unchanged — checked bit for bit against the code without the padding, for XPHM (with its internal multibanding on), Pv2, Pv2_NRTidal, XAS and D, on a heavily truncated and on a standard multibanded domain — and a new regression test checks that XP_NRTidalv3 now gives the same waveform on differently truncated grids.

Issue #420, opened 2026-09-15
Importance weights are biased for samples entering the prior after synthetic phase


nihargupte-ph:
## Summary

If the prior is updated **after** `Result.sample_synthetic_phase` so that previously out-of-prior samples fall within the new prior, `importance_sample` assigns those samples weights that are inconsistent with the proposal. `dingo_pipe` is not affected, since it calls `update_prior` before the synthetic phase, but the Python API allows the other order.

## What happens

1. `sample_synthetic_phase` only draws a phase for samples within the prior at that time. Out-of-prior samples get the placeholder `phase = 0` and `log_prob = nan` ([dingo/gw/result.py#L604](https://github.com/dingo-gw/dingo/blob/f6467144/dingo/gw/result.py#L604)). This discards the full stored proposal density, including the network's `log q(theta)`.
2. `update_prior` is called with a wider prior, so some of these samples now have a finite prior.
3. `importance_sample` selects valid samples from the prior alone, `np.isfinite(log_prior + delta_log_prob_target)` ([dingo/core/result.py#L332](https://github.com/dingo-gw/dingo/blob/f6467144/dingo/core/result.py#L332)). It evaluates the likelihood at the placeholder `phase = 0`.
4. `_calculate_evidence` uses `nan_to_num(log_prob)` ([dingo/core/result.py#L400](https://github.com/dingo-gw/dingo/blob/f6467144/dingo/core/result.py#L400)), i.e. `q = 1`. The weight becomes `pi(theta) L(theta, phase=0)`, with no proposal density in the denominator.

The phase of these samples was never drawn, and their proposal density is unknown. Their weights are off by the factor `q(theta) q(phase | theta)`, which biases the evidence and the weights. This is a bias, not just extra variance.

Samples that stay outside the prior are unaffected, since `log_prior = -inf` gives them zero weight regardless.

## Possible fixes

- **Minimal:** `importance_sample` also treats samples with non-finite `log_prob` as invalid (zero weight). This is correct, but those samples are wasted.
- **Complete:** when the prior changes so that new samples fall within it, draw the synthetic phase (and calibration parameters, if sampled) for those samples, so they get a proper proposal density. This likely fits with moving the evolving prior and importance-sampling preparation out of `Result`, as planned in #389.

With a cached synthetic-phase log likelihood (upcoming PR), `importance_sample(use_cached_log_likelihood=True)` raises an error in this situation instead, because these samples have no cached value.

PR #422, opened 2026-09-15
Cache the synthetic-phase log likelihood for importance sampling

nihargupte-ph:
**Stacked on #421** (base: `calibration-prior-factor`). Retarget to `main` once #421 is merged.

## Summary

The exact synthetic phase already evaluates the likelihood as a function of phase from **one waveform call per sample**. Importance sampling then evaluates the likelihood again at the drawn phase, i.e. a second waveform call. This PR caches the log likelihood at the drawn phase during the synthetic phase and reuses it in importance sampling, which removes the second waveform evaluation.

## How it works

- **Likelihood** (`dingo/gw/likelihood.py`): the mode-decomposed phase grid is split into
  - `phase_grid_terms(theta)`: one waveform call; returns the per-mode inner products `(d | mu_m)` and the cross terms `(mu_m | mu_n)`;
  - `log_likelihood_from_phase_grid_terms(terms, phases)`: log L at any phases from those terms, without a waveform call.

  `log_likelihood_phase_grid` is the composition of the two, with unchanged results.
- **`SyntheticPhaseFactor(cache_log_likelihood=False)`**: keeps the terms and, after the draw, evaluates log L **exactly at the drawn phase**, emitted as a `log_likelihood` column. Exact mode only; combining it with `approximation_22_mode` raises an error.
  - The phase is drawn from the grid distribution interpolated between grid points, so it generally lies between grid points and its likelihood is not one of the grid values.
  - Snapping the draws to grid points would make the proposal discrete, inconsistent with the continuous interpolated `log_prob`.
  - Interpolating the stored grid of log likelihoods would only be approximate.
- **`Result.sample_proposal_extensions`** writes the column (NaN outside the prior). Calibration is drawn earlier in the same chain (#421), so the cached value includes the drawn calibration curve. Drawing calibration without the synthetic phase drops a stale cached column.
- **`importance_sample(use_cached_log_likelihood=False)`**: when true, uses the cached values instead of evaluating the likelihood.
  - Raises if the column is missing or a marginalization is requested.
  - Raises if any sample within the prior lacks a finite cached value, e.g. if the prior was widened after the synthetic phase. That case would otherwise be biased, see #420.
- **`dingo_pipe`**: `cache_log_likelihood: True` in `PhaseRecoveryDefault`. The pipeline disables it (with a log message) for the (2, 2) approximation or any marginalized likelihood.

## Design decision: how the cached column leaves the chain

`ChainComposer` drops the extra columns a step emits beyond its parameters (e.g. the detector times GNPE recomputes). Previously the only exception was `isinstance(step, TargetCorrection)`, which is how `delta_log_prob_target` reaches importance sampling. `SyntheticPhaseFactor` is a `Factor`, so its `log_likelihood` column was dropped.

This PR replaces that special case with one rule: **a column is kept iff the step lists it in `annotations`**.
- `TargetCorrection.annotations` returns its `produces`, so `delta_log_prob_target` behaves as before.
- `Factor.annotations` defaults to empty, and `SyntheticPhaseFactor` declares `["log_likelihood"]` when caching.

The cached log likelihood and `delta_log_prob_target` thus take the same path.

Rejected alternative: a separate `TargetCorrection` step after the phase factor. It would need either a second waveform call (defeating the purpose) or passing the per-sample complex mode terms through the chain as extra columns.

## Validation

S240527en (SEOBNRv5PHM, production proposal, 240 samples, calibration sampled, IS only):
- All 231 valid samples used cached values; 0 likelihood evaluations.
- Cached vs direct `log_likelihood` (80 samples): median 3e-5, max 1.5e-4 nats. The residual is the DFT phase-decomposition path vs the direct polarization call; the chain's float32 parameters contribute ≤ 7e-6.
- logZ −8601.24 ± 0.14 vs −8601.31 ± 0.12 (`main`).

Full-scale reruns of the three events from #381 (`main` vs #421 vs this PR, importance sampling only, from the production proposals) will be attached here.

## Tests

- Terms reproduce the direct likelihood at off-grid phases, with and without calibration.
- The factor emits the correct cached value; the draw and `log_prob` are identical with and without caching; the column survives the chain; the (2, 2) combination raises.
- `importance_sample` cache: uses cached values without evaluating the likelihood; raises for missing coverage, a missing column, and marginalization; ignores the cache by default.

https://claude.ai/code/session_01373LokcH2Tb38prnpatM2C


nihargupte-ph:
## Full-scale validation: S240527en, S240618ah, S250118t

Importance sampling was rerun for the three events from #381, starting from the GWTC-5 production proposal samples. Each event used 10 jobs × 50k samples, 32 CPUs per job, SEOBNRv5PHM, and calibration sampled. Four series are compared:

- **#422**: this PR, with `cache_log_likelihood: True`
- **#421**: calibration and synthetic phase in one chain, no cache
- **`main`** (f6467144)
- **production**: the original GWTC-5 importance-sampling result, which ran older dingo code

| Event | Series | log Z | n_eff | Efficiency | Mean time per IS job |
|---|---|---|---|---|---|
| S240527en | #422 | −8601.193 ± 0.005 | 36,088 | 7.22% | **467 s** |
| | #421 | −8601.194 ± 0.005 | 34,628 | 6.93% | 733 s |
| | main | −8601.192 ± 0.005 | 31,850 | 6.37% | 715 s |
| | production | −8601.192 ± 0.006 | 29,796 | 5.96% | 1692 s |
| S240618ah | #422 | −5737.540 ± 0.003 | 114,957 | 22.99% | **411 s** |
| | #421 | −5737.539 ± 0.003 | 109,108 | 21.82% | 649 s |
| | main | −5737.541 ± 0.003 | 105,220 | 21.04% | 634 s |
| | production | −5737.540 ± 0.003 | 105,010 | 21.00% | 1585 s |
| S250118t | #422 | −8596.527 ± 0.003 | 115,646 | 23.13% | **472 s** |
| | #421 | −8596.529 ± 0.003 | 117,116 | 23.42% | 733 s |
| | main | −8596.528 ± 0.003 | 117,229 | 23.45% | 715 s |
| | production | −8596.528 ± 0.003 | 116,563 | 23.31% | 1710 s |

### Takeaways

- **Results are unchanged.** log Z agrees within ~1σ across all four series for every event. The largest Jensen–Shannon divergence over all 15 parameters is 0.47×10⁻³ bit, compared with the 2×10⁻³ bit indistinguishability threshold.
- **The likelihood stage is removed.** With the cache, importance sampling uses the stored log likelihoods for every valid sample: 0 likelihood evaluations, versus 240–263 s per job without the cache. The synthetic phase itself costs the same as in #421.
- **Each job is ~1.55× faster**: 1.56–1.58× faster than #421 and 1.52–1.54× faster than `main`.
- **#421 changes efficiency only slightly.** Conditioning the synthetic phase on the drawn calibration raises efficiency relative to `main` for S240527en (+9%) and S240618ah (+4%), and leaves S250118t unchanged. #421 and #422 draw phases with the same algorithm, so the differences in efficiency between them are sampling noise.
- **Production timings are not comparable.** Production ran older dingo code with a slower synthetic phase (~1300–1400 s per job), which accounts for its longer job times.

Per-event figures below: posteriors, the full JSD table, log Z / n_eff, and timing split by stage.

validation_summary_S250118t
validation_summary_S240618ah
validation_summary_S240527en

lorenzopompili00:
Thanks @nihargupte-ph this looks great! I have done some more testing and things look good, but there are some code paths where this currently leads to some error and the caching should be disabled. 

Here is a Claude-assisted summary.

===============================

The cached value comes from `generate_hplus_hcross_m`, and the direct one from `generate_hplus_hcross`.  Only the DFT paths and the SEOBNRv5 conditioning mimic reproduce `generate_hplus_hcross`. The other paths have known mismatches, already listed in the tolerances of `tests/gw/waveform_generator/test_wfg_m.py` (up to 2e-2 for IMRPhenomXPHM, 7e-4 for SEOBNRv4PHM, 3e-3 for NRSur7dq4). There are three causes:

- **Individual-mode XPHM** computes the orbital angular momentum outside the model.
- **Time-domain models** have each mode tapered and Fourier-transformed on its own, while the direct call conditions the summed polarizations.
- **LAL frequency-domain models on a multibanded grid** compute the modes on the base grid and average them into bands, while the direct call evaluates the waveform at the band frequencies.

Where the two differ, the difference used to cost only proposal efficiency. With caching it goes into the weights, and nothing reports it. They do differ on several paths; the mismatches are already visible in the tolerances of `test_wfg_m.py`.

Cached minus direct log L near an injection, median / max |Δ log L|:

| Waveform path | SNR 20 | SNR 80 |
|---|---|---|
| DFT (IMRPhenomXPHM, SEOBNRv5PHM), SEOBNRv5HM | below 4e-4 / 1e-3 | below 3e-3 / 9e-3 |
| IMRPhenomXPHM, `use_dft_phase_decomposition: False` | 0.011 / 0.06 | 0.06 / 0.23 |
| IMRPhenomXPHM, multibanded without `use_base_domain` | 0.17 / 0.42 | 0.72 / 1.6 |
| SEOBNRv4PHM (time-domain modes) | 0.05 / 0.10 | 0.14 / 0.19 |
| NRSur7dq4 (time-domain modes) | 0.14 / 0.32 | 0.87 / 1.2 |
| SEOBNRv5EHM at e = 0 (generic new-interface time-domain branch) | 0.06 / 0.11 | 0.30 / 0.51 |

The error changes from sample to sample, so the posterior shifts, not just log Z.

**Suggestion:** cache only when the m-components are rebuilt from the model's own polarizations on the grid the likelihood uses. That is what the DFT branch of `generate_hplus_hcross_m` does, and where caching is exact up to numerical noise.

- **How.** The waveform generator could set a flag in that branch; don't decide by approximant name. `SyntheticPhaseFactor` would skip the cache, with a warning, when the flag is not set. `dingo_pipe` then evaluates the likelihood directly, as before.
- **What it costs.** SEOBNRv5HM and SEOBNRv5PHM with individual modes are accurate but would lose the speed-up.
- **The multibanded caveat.** For LAL models on a multibanded grid without `use_base_domain`, the DFT branch averages the modes into bands, while `generate_hplus_hcross` evaluates at the band frequencies, and the two differ by a decimation error. However one could also directly evaluate the phase grid at the MFD frequencies instead, in which case the m-components would sum up exactly to what `generate_hplus_hcross` returns. **I will open a PR for this.**

nihargupte-ph:
Thanks, I've updated it to have a check that will only use the DFT path when it's exact. 

One question for @lorenzopompili00, I know it's not currently implemented in the code, but does SEOBNRv5EHM also have the exact DFT the same as SEOBNRv5PHM?

nihargupte-ph:
Thanks @lorenzopompili00, agreed. Both PRs are rebased onto `main` (incl. #424), and caching is now restricted to the DFT phase decomposition (a5b65693):

- **`WaveformGenerator.uses_dft_phase_decomposition`**: a property on both generator classes saying whether `generate_hplus_hcross_m` takes the DFT branch. The branches themselves use it, so the condition lives in one place rather than in a list of approximant names.
- **`dingo_pipe`** builds a waveform generator from the network settings (plus any `synthetic_phase: use_dft_phase_decomposition` override) and disables the cache if that property is False, as it already does for the (2, 2) approximation and marginalized likelihoods. Outside the pipe, `cache_log_likelihood` stays opt-in.

With #424 the DFT m-components sum to `generate_hplus_hcross` on multibanded domains too (max relative difference 7e-16 for IMRPhenomXPHM, 1.2e-6 for SEOBNRv5PHM, vs ~3e-3 with DFT off), so the cache is kept for both regardless of the grid. The three validation events use SEOBNRv5PHM with DFT, so the results above are unaffected.

lorenzopompili00:
> One question for @lorenzopompili00, I know it's not currently implemented in the code, but does SEOBNRv5EHM also have the exact DFT the same as SEOBNRv5PHM?

No, this currently does not work exactly for SEOBNRv5EHM. The issue is with the SEOBNRv5EHM tapering, which has a slight dependence on the phase (the taper's length is set by where the signal's first peaks fall, so it shifts with the phase and differs between h_+ and h_x). So the waveform is not exactly a sum of m-components times e^(−i m phi).

However for aligned spins there is not as big of a difference in speed between the mode path that is currently used and the one of https://github.com/dingo-gw/dingo/pull/381

dingo/gw/inference/steps.py
```diff
@@ -124,21 +153,28 @@ def sample_and_log_prob(self, num_samples, context, given=None):
         n = len(reference)
         logger.info(f"Estimating synthetic phase for {n} samples.")
         t0 = time.time()
-        phases, phase_posterior = self._phase_profile(given, context)
+        phases, phase_posterior, terms = self._phase_profile(given, context)
```
lorenzopompili00, line 156: I would add a check here as well to make sure that direct call through the Python API does not cache silently when not intended (never reached by `dingo_pipe`). The two cache tests in `test_synthetic_phase.py` then need `_MockLikelihood.waveform_generator = SimpleNamespace(uses_dft_phase_decomposition=True, approximant_str="mock")`.

```suggestion
        if self.cache_log_likelihood:
            # The cache matches a direct likelihood call only with the DFT decomposition.
            waveform_generator = context.likelihood(
                use_base_domain=self.use_base_domain, wfg_updates=self.wfg_updates
            ).waveform_generator
            if not waveform_generator.uses_dft_phase_decomposition:
                raise ValueError(
                    "cache_log_likelihood requires the DFT phase decomposition, which "
                    f"{waveform_generator.approximant_str} does not use here."
                )
        phases, phase_posterior, terms = self._phase_profile(given, context)
```

docs/source/result.md
```diff
@@ -44,6 +44,9 @@ n_grid
 uniform_weight
 : Base probability level to add to ensure mass coverage.
 
+cache_log_likelihood
+: Whether to also store the log likelihood at the sampled $\phi_c$ in the `log_likelihood` column (exact mode only), so that `importance_sample(use_cached_log_likelihood=True)` does not need to generate the waveform again. The sampled $\phi_c$ is drawn from the grid distribution interpolated between grid points, so its likelihood is evaluated exactly from the cached mode inner products rather than read off the grid. `dingo_pipe` enables this by default when the likelihood is not marginalized.
```
lorenzopompili00, line 48: I would add for completeness
```suggestion
: Whether to also store the log likelihood at the sampled $\phi_c$ in the `log_likelihood` column (exact mode only), so that `importance_sample(use_cached_log_likelihood=True)` does not need to generate the waveform again. The sampled $\phi_c$ is drawn from the grid distribution interpolated between grid points, so its likelihood is evaluated exactly from the cached mode inner products rather than read off the grid. The cached value matches a direct evaluation when the m-components obtained from `generate_hplus_hcross_m` sum to exactly the waveform from `generate_hplus_hcross`. This is guaranteed by construction for the DFT phase decomposition, so `dingo_pipe` enables caching by default only when the likelihood is not marginalized and the DFT phase decomposition is used.
```

dingo/gw/inference/steps.py
```diff
@@ -102,7 +115,16 @@ def __init__(
             Overrides for the waveform generator settings stored with the network,
             e.g. `use_dft_phase_decomposition` (passed on to
             `SamplerContext.likelihood`).
+        cache_log_likelihood : bool, default False
+            Also emit the log likelihood at the drawn phase as the column
+            `log_likelihood`, for reuse by importance sampling. Requires the exact
+            mode (`approximation_22_mode=False`).
```
lorenzopompili00, line 121:
```suggestion
            mode (`approximation_22_mode=False`) and the DFT phase decomposition.
```

dingo/core/result.py
```diff
@@ -299,6 +310,17 @@ def importance_sample(self, num_processes: int = 1, **likelihood_kwargs):
                 "it is necessary to train an unconditional flow based on the existing "
                 "samples. This can then be sampled with log probability."
             )
+        if use_cached_log_likelihood:
+            if "log_likelihood" not in self.samples:
+                raise KeyError(
+                    "use_cached_log_likelihood=True requires log likelihoods stored "
+                    "in the samples."
+                )
+            if any(likelihood_kwargs.values()):
```
lorenzopompili00, line 319: The saved values can't be combined with marginalization, and this line checks for that. But it treats an empty settings dict `{}` as "no marginalization", while `_build_likelihood` treats `{}` as "marginalization on, with default settings". So that one input slips through. It's a rare case; the fix makes both use the same test.

```suggestion
            if any(v is not None for v in likelihood_kwargs.values()):
```

and in `pipe/importance_sampling.py` L274:

```
                or any(v is not None for v in likelihood_kwargs.values())
```

dingo/gw/result.py
```diff
@@ -508,6 +513,9 @@ def _calibration_steps(self, calibration_sampling_kwargs: dict) -> list:
             correction_type_dict = correction_type
         else:
             raise ValueError(f"{correction_type} not understood")
+        if "log_likelihood" in self.samples:
+            # A stored log likelihood does not include the new calibration.
+            self.samples = self.samples.drop(columns="log_likelihood")
```
lorenzopompili00, line 518: An edge case (Python API only; `dingo_pipe` is fine): the cached-values check only tests that a `log_likelihood` column exists, and a normal importance-sampling run writes to that same column. So values left over from an earlier run can be reused by mistake:

```python
result.importance_sample(phase_marginalization_kwargs={...})   # writes log_likelihood (phase-marginalized)
result.sample_proposal_extensions(synthetic_phase_kwargs={...})  # new phases, but cache_log_likelihood not set
result.importance_sample(use_cached_log_likelihood=True)        # silently reuses the old, marginalized values
```

Today the column is only deleted when calibration is redrawn. Deleting it whenever phases or calibration are drawn makes the last call fail with the existing `KeyError` instead, pointing to the missing `cache_log_likelihood=True`. Add at the start of `sample_proposal_extensions`, after L432:

```python
        if "log_likelihood" in self.samples:
            self.samples = self.samples.drop(columns="log_likelihood")
```

dingo/core/result.py
lorenzopompili00, line 214: A `log_likelihood` column cached on the old data would otherwise be reused by `importance_sample(use_cached_log_likelihood=True)` with the new data (same idea as the drop in `_calibration_steps`)
```suggestion
        if self.samples is not None and "log_likelihood" in self.samples:
            self.samples = self.samples.drop(columns="log_likelihood")
```

dingo/core/inference/steps.py
```diff
@@ -161,11 +161,16 @@ class Factor(ABC):
         Whether the factor draws new samples (the default) or is a point mass or
         fixed table that is run once. The chain's sample counts go to the steps
         that draw, one each; an int `num_samples` is the count for the first.
+    annotations : list[str]
+        Columns the factor emits beyond its `parameters` that are kept in the chain
+        output for importance sampling, such as a cached log likelihood. Other side
+        channels are dropped. Target corrections declare their columns the same way.
```
lorenzopompili00, line 167: Each step describes its columns with three lists: `parameters` (what it samples), `produces` (every column it writes) and, new here, `annotations` (extra columns to keep in the output). Two places only look at `produces`:
- the chain's check at construction, which catches a step overwriting an earlier column;
- `sample_proposal_extensions`, which copies the `produces` columns into the samples.

The base `Factor.produces` returns only `parameters`. `SyntheticPhaseFactor` works because it overrides `produces` to add its annotations, but a new factor that only sets `annotations = ["x"]`, as this docstring suggests, would have `x` missed by both.

**Suggestion:** make `Factor.produces` include `annotations` by default, so that setting `annotations` is enough. That takes three small changes:

- this docstring:

```suggestion
    annotations : list[str]
        Columns the factor emits beyond its `parameters` that are kept in the chain
        output for importance sampling, such as a cached log likelihood. They are part
        of `produces`; other side channels are dropped.
```

- `Factor.produces`, L177–181 (`"""The emitted columns: ...`):

```python
        """The emitted columns: `parameters` and `annotations`, plus any side
        channels (overridden then). A side channel is an intermediate that later
        steps may read, such as the detector times a GNPE network recomputes; unlike
        an annotation, it is not part of the chain's output."""
        return self.parameters + self.annotations
```

- Remove the now-redundant override in `dingo/gw/inference/steps.py` L139–143 (`@property` / `def produces`):

Could you also check the documentation of `annotations` again? Some places still say that every column beyond `parameters` is dropped, e.g. the `Step` protocol docstring in `core/inference/steps.py` and `docs/source/sampling_chains.md`.

lorenzopompili00 (changes requested):

dingo/core/inference/steps.py
lorenzopompili00: I would also like to have @stephengreen's opinion on the addition of `annotations` since this changes the main `Factor` base class and I am not as familiar with its logic.

lorenzopompili00:
Thank you @nihargupte-ph! The fix in https://github.com/dingo-gw/dingo/pull/422/commits/a5b656930910ee23655420ab0b1322fdf9829484 seems to work well. I left a few more minor comments (mostly covering some corner cases beyond standard use with `dingo_pipe`). The more substantial comment is on the addition of `annotations` to the main `Factor` base class where I might need to understand a bit better the logic.

lorenzopompili00:
Just a reminder: the `co_rotate_spins` option in https://github.com/dingo-gw/dingo/pull/419 would also allow to cache the synthetic-phase log likelihood for importance sampling for BNS models with only the co-precessing (2,|2|) modes.

This should require only a few small edits. We can edit this PR or https://github.com/dingo-gw/dingo/pull/419 depending on which one is merged first.

PR #424, opened 2026-09-15
Evaluate the DFT phase grid at the multibanded frequencies

lorenzopompili00:
## Summary

On a `MultibandedFrequencyDomain`, the DFT branch of `generate_hplus_hcross_m` (#381) evaluated the phase-grid polarizations on the base domain and then averaged the m-components into MFD bins. For frequency-domain approximants, `generate_hplus_hcross` evaluates the waveform at the MFD frequencies instead (`SimInspiralChooseFDWaveformSequence`). As a result, the m-components did not sum exactly to that waveform, which made the synthetic-phase likelihood differ slightly from the direct one.

This PR evaluates the phase grid at the MFD frequencies instead. The base-domain detour is only needed by the individual-mode route, because `SimInspiralChooseFDModes` does not support multibanding. The DFT route only needs polarizations, which do support it.

## Effect

For IMRPhenomXPHM on an MFD (20–1038 Hz), comparing `main` with this branch:

| | `main` | This PR |
|---|---|---|
| m-components summed vs `generate_hplus_hcross`, mismatch | 7.7e-6 | 2e-16 |
| Phase-grid vs direct log likelihood, SNR 80, median / max | 0.63 / 1.5 | 1e-12 / 2e-12 |
| `generate_hplus_hcross_m` time per call | 0.35 s | 0.03 s |

This affects exact synthetic phase with multibanded networks of frequency-domain approximants when the likelihood is evaluated on the MFD (`use_base_domain=False`, the `Result` default). Today the mismatch only lowers slightly importance-sampling efficiency. It would bias the weights once the synthetic-phase log likelihood is cached (#422). Time-domain approximants (e.g. SEOBNRv5PHM) are generated on the base domain and decimated on both routes, so they are unchanged.

## Tests

- **New:** `test_dft_reconstructs_phase_shift_on_mfd` checks that the m-components sum to `generate_hplus_hcross` at off-grid phase shifts, to round-off. It fails on `main` with a mismatch of 7e-7.
- **Changed:** `test_decimation_m` compares each m-component computed on the MFD against the decimated base-domain component.
  - Before this PR both sides were decimated, so they matched exactly.
  - For IMRPhenomXPHM they now differ by the decimation error. That error is small for the full waveform, still held to 1e-4 in `test_decimation`.
  - The per-component mismatch is normalized by each component's own power, so it becomes large for components that carry almost no power or oscillate fastest (high |m|). Over 100 prior draws the largest per-component mismatch had a median of 6e-5 and a maximum of 1.6e-2, so the IMRPhenomXPHM bound in this test is now 5e-2. The time-domain approximants keep their exact tolerance.
- **Run:** `test_wfg_mfd.py`, `test_wfg_m.py`, `test_mfd.py`, `test_likelihood_phase_grid.py`, `test_likelihood.py` and `test_synthetic_phase.py` all pass.


nihargupte-ph (approved):
LGTM

PR #430, opened 2026-09-18
Determine multibanded domains from heterodyned waveforms (DINGO-BNS)

stephengreen:
dingo_generate_multibanded_domain and dingo_evaluate_multibanded_domain now
honor `phase_heterodyning` in the dataset compression settings: the bands are
determined from, and the mismatch evaluated on, heterodyned waveforms, which
is what a chirp-mass-conditioned network sees. A new
`--chirp_mass_proxy_offset` heterodynes at the chirp mass plus the largest
offset of the training kernel, the worst case for the residual oscillation.


stephengreen:
Two follow-up commits (37a1d3eb, 7371cbfd): the proxy offset now alternates in sign by row, and the mismatch target applies to the worse of the two sides.

**Why.** The residual after heterodyning at the proxy is the post-Newtonian remainder plus an offset term whose sign flips with the offset, so the two sides of the kernel decimate differently and a domain determined from one side alone can fail the other. Checked two ways.

Toy BNS domain (20 to 256 Hz, XP_NRTidalv3, target 1e-4, offset 0.005), median mismatch on fresh waveforms heterodyned at each edge of the kernel:

| domain built from | bins | proxy at +δ | proxy at −δ |
|---|---|---|---|
| +δ only | 997 | 9.2e-5 | 1.45e-4 |
| −δ only | 891 | 5.2e-4 | 1.0e-4 |
| alternating | 1360 | 8.0e-5 | 1.0e-4 |

Neither side is uniformly worse there, and each one-sided domain misses the opposite side (by 1.5× and 5×).

The example settings from #431 (20 to 2048 Hz, target 1e-5, offset 0.005, 100 samples), with the worse-side target:

| domain built from | bins | proxy at +δ | proxy at −δ |
|---|---|---|---|
| +δ only (as shipped) | 3758 | 9.7e-6 | 3.6e-7 |
| −δ only | 2588 | 5.3e-4 | 9.6e-6 |
| alternating | 3730 | 1.03e-5 | 3.7e-7 |

Here +δ is the hard side everywhere, so the shipped one-sided domain happened to be the safe choice; a −δ-only domain would have been 50× off. With the mixture median as the target the alternating domain was 30% too coarse on the hard side (1.3e-5), which is what the second commit fixes. Max's strict check on the alternating domain, waveforms generated pointwise on the multibanded domain against base-domain waveforms decimated, gives a maximum mismatch of 8e-7 at +δ and 2e-7 at −δ.

Regenerating the example domain with the new code moves the nodes by under 1 Hz and keeps 8 bands and about 3730 bins, so the shipped `waveform_dataset_settings_mfd.yaml` is left as is. The new test runs the tool on the toy domain and asserts that the alternating domain meets the target on both edges and that a one-sided domain is worse on the edge it never saw (6 s).

stephengreen:
@max-dax @nihargupte-ph I'm merging this in the interest of time, but it would be great if one of you could look at the updated version of the MFD generator. I'm not super familiar with this function.

PR #436, opened 2026-09-21
Compute the waveform time-shift phase in double precision

lorenzopompili00:
## Summary

`UniformFrequencyDomain` stores its frequencies as float32. After generating a waveform, dingo multiplies it by `exp(-2πi f dt)` to undo the epoch of the LAL / gwsignal output, with `dt` ≈ 0.1 s. The prefactor `-1j * 2 * np.pi * dt` is a Python scalar, so NumPy keeps the float32 precision of the frequency array and computes the phase factor in complex64. Two errors follow:

- Rounding `2π dt` to single precision changes `dt` by up to `dt · 2⁻²⁴` ≈ 6 ns, which shifts the waveform in time.
- Rounding the phase itself (up to ~700 rad at 1 kHz) adds ~2×10⁻⁶ rad of noise.

This PR casts the frequencies to float64 before computing the phase, in the three places that apply this shift:

- `WaveformGenerator.generate_FD_waveform` (LAL route);
- `NewInterfaceWaveformGenerator.generate_FD_waveform` (gwsignal route, e.g. SEOBNRv5PHM);
- `wfg_utils.td_modes_to_fd_modes` (individual-mode route of time-domain approximants).

The domain itself is left in float32, since changing its dtype would touch datasets and network inputs. The DFT phase decomposition (#381) was not affected, because pyseobnr applies its shift in float64.

## Effect

Direct waveform (`generate_hplus_hcross`) on 20–1024 Hz. For each case, the table gives the fitted time offset and the residual phase scatter against a reference, weighted by the design PSD:

| | `main` | This PR |
|---|---|---|
| IMRPhenomXPHM, Δf = 1/8 Hz | +1.80 ns, 1.7e-6 rad | 0.00 ns, 1e-17 rad |
| SEOBNRv5PHM, f_start 20 Hz, Δf = 1/8 Hz | −1.91 ns, 2.0e-6 rad | −0.04 ns, 7e-15 rad |
| SEOBNRv5PHM, f_start 9 Hz, antisymmetric modes, Δf = 1/32 Hz | +3.57 ns, 1.8e-6 rad | −0.20 ns, 7e-15 rad |

For IMRPhenomXPHM the reference is the same computation with float64 frequencies. For SEOBNRv5PHM it is pyseobnr's conditioned time-domain output, FFT'd and shifted with the model's exact floating-point start time. The sub-ns offset left after the fix comes from storing that start time as a `LIGOTimeGPS`, which rounds it to the nanosecond. The DFT route has offsets of the same size (−0.30 and +0.05 ns).

Log likelihood error of the SEOBNRv5PHM direct waveform against the same reference, for parameters near an injection (median / max over 6 draws):

| | `main` | This PR |
|---|---|---|
| f_start 20 Hz, SNR 20 | 3.2e-4 / 7.1e-4 | 2.4e-5 / 5.1e-5 |
| f_start 20 Hz, SNR 80 | 1.1e-3 / 7.0e-3 | 4.7e-5 / 3.5e-4 |
| f_start 9 Hz, SNR 20 | 3.4e-4 / 4.8e-4 | 2.6e-5 / 5.4e-5 |
| f_start 9 Hz, SNR 80 | 3.0e-3 / 1.0e-2 | 7.0e-4 / 1.5e-3 |

The SEOBNRv5PHM individual-mode route had the same offsets as the direct route, and after the fix it is within 0.05 ns of it. It keeps ~1e-6 rad of phase scatter, which is genuine and comes from conditioning each mode separately. These errors are small, but they were the main disagreement between the direct and DFT routes for SEOBNRv5PHM.

## Tests

- **New:** `test_time_shift_in_double_precision` checks that the direct waveform and the individual-mode m-components are the same, to 1e-12, whether the domain's frequencies are float32 or float64. It fails on `main` for both IMRPhenomXPHM and SEOBNRv5PHM.
- **Tightened:** `test_dft_reconstructs_phase_shift` compares the summed DFT m-components with `generate_hplus_hcross` at off-grid phase shifts.
  - For SEOBNRv5PHM the mismatch goes from 2.9e-12 to 6.2e-14 and the amplitude error from 3.4e-9 to 1.7e-14. IMRPhenomXPHM stays at 2e-16 and 1e-15.
  - The bounds are now 1e-12 for both, down from 1e-9 (mismatch) and 1e-6 (amplitude). The SEOBNRv5PHM case fails on `main`.
- **Docstring:** `dft_vs_standard_tolerance` no longer mentions a ~1 ns epoch-rounding offset. The DFT-vs-individual-mode mismatch it bounds is unchanged (1.8e-5 for IMRPhenomXPHM, 6.9e-7 for SEOBNRv5PHM).
- **Run:** all 104 tests pass in `tests/gw/waveform_generator/`, `test_likelihood*.py`, `test_injection*.py`, `test_synthetic_phase.py`, `test_mfd.py`, `test_ufd.py` and `test_gw_result*.py`.


stephengreen:
Nice catch. The fix looks good to me.

One question: should we make the same change in `BaseFrequencyDomain.time_translate_data()`? It has the same pattern, `2π · einsum(dt, f)` with the float32 grid. At the moment the numpy branch comes out float64 only because `np.einsum` promotes a float64 `dt` (the detector times in `ProjectOntoDetectors`, or the Python-float post-trigger shift in `data_generation.py`), and the torch branch is float32 by construction (float32 grid times float32 `dt`). The torch branch is currently only used for the GNPE proxy shifts at inference, where `dt` is a few ms, so the phase is small and it doesn't matter. But a float32 `dt` array in numpy, or moving the detector projection onto the GPU, would silently drop the training shifts (`dt` up to ~0.1 s) to single precision, with the same few-ns / 1e-5 rad errors as here.

The fix would have the same shape as in `factor_fiducial_waveform`: compute the phase in float64 in both branches and cast the phasor back to the data's dtype. In numpy that would also remove an incidental upcast of the strain to complex128 in the training transforms. Happy to do it in a follow-up PR if you'd rather keep this one minimal.

lorenzopompili00:
> Nice catch. The fix looks good to me.
> 
> One question: should we make the same change in `BaseFrequencyDomain.time_translate_data()`? It has the same pattern, `2π · einsum(dt, f)` with the float32 grid. At the moment the numpy branch comes out float64 only because `np.einsum` promotes a float64 `dt` (the detector times in `ProjectOntoDetectors`, or the Python-float post-trigger shift in `data_generation.py`), and the torch branch is float32 by construction (float32 grid times float32 `dt`). The torch branch is currently only used for the GNPE proxy shifts at inference, where `dt` is a few ms, so the phase is small and it doesn't matter. But a float32 `dt` array in numpy, or moving the detector projection onto the GPU, would silently drop the training shifts (`dt` up to ~0.1 s) to single precision, with the same few-ns / 1e-5 rad errors as here.
> 
> The fix would have the same shape as in `factor_fiducial_waveform`: compute the phase in float64 in both branches and cast the phasor back to the data's dtype. In numpy that would also remove an incidental upcast of the strain to complex128 in the training transforms. Happy to do it in a follow-up PR if you'd rather keep this one minimal.

Thanks @stephengreen! Yes I agree that operation could also be enforced to use float64. I think we can keep this PR minimal and do it in a separate one. Happy for you to take it, or I can do it myself, whichever you prefer.

